PR #6, opened 2024-02-20
Added tests for backend part

dmitryk-dk:
Added tests to the backend part of the plugin

### DIFF
--- a/pkg/plugin/response_test.go
+++ b/pkg/plugin/response_test.go
@@ -64,15 +64,7 @@ func Test_parseStreamResponse(t *testing.T) {
 			},
 		},
 		{
-			name:     "incorrect time in the response",
-			reader:   f,
-			response: `{"_time":"acdf"}`,
-			want: func() backend.DataResponse {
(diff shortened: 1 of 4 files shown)


Loori-R (approved):
LGTM!

pkg/plugin/response_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/grafana/grafana-plugin-sdk-go/data"
+)
+
+func Test_parseStreamResponse(t *testing.T) {
+	f := func(remoteResp string) io.Reader {
+		return io.NopCloser(bytes.NewBuffer([]byte(remoteResp)))
+	}
+	tests := []struct {
+		name     string
+		reader   func(remoteResp string) io.Reader
```
hagen1778, line 21: `reader` seems identical in every test case. Do you need it?

dmitryk-dk: Removed it as a param of the test struct

hagen1778 (approved):
lgtm

Issue #12, opened 2024-03-10
Time range ignored


TertiumOrganum1:
Sometimes source refused to take time range into account, especially when specify particular range boundaries (not 'now'). Behavior remain even after ctrl+shift+r.
I found a correlation - this happens when I specify time values as From and To and add non-msg field in query:
_stream:{service=gateway,level=error,status=408} ingest_timestamp:*

![image](https://github.com/VictoriaMetrics/victorialogs-datasource/assets/51286827/bbb203b9-bb9c-4390-9d10-5ebd7c4fc62a)

Loori-R:
Could you wrap the query in parentheses and check if the time filter is applied correctly to ensure the issue is related to https://github.com/VictoriaMetrics/VictoriaMetrics/issues/5920:
`(_stream:{service=gateway,level=error,status=408} ingest_timestamp:*)`

TertiumOrganum1:
this behaviour is not stable, but I'll try next time

вс, 10 мар. 2024 г., 16:53 Yury Molodov ***@***.***>:

dmitryk-dk:
This bug has been fixed in [v0.2.0 release](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.2.0). Closing the issue as fixed then.

Issue #10, opened 2024-03-04
First request returns empty response


dmitryk-dk:
Sometimes, when you run the Grafana explore panel, enter the request, and the request executed in the response returns the empty data. But the VictoriaLogs has data in the storage.

It happens from time to time.

dmitryk-dk:
This bug has been fixed in [v0.2.0 release](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.2.0). Closing the issue as fixed then.

Issue #9, opened 2024-03-04
No logs limits


dmitryk-dk:
VictoriaLogs starts to support limiting the number of returned log entries from v0.5.0. 
https://docs.victoriametrics.com/victorialogs/changelog/#v050
Some of the users can use the previous versions. The previous version of the VictoriaLogs can return more than 11000 lines.
Such a number of lines slows down the operation of the interface, and the user interface becomes inconvenient.
<img width="1026" alt="Screenshot 2024-03-04 at 11 19 40" src="https://github.com/VictoriaMetrics/victorialogs-datasource/assets/12189570/8746f442-8dc7-42e4-b3d2-b15ba627ea32">

I think we should limit the number of lines in the frontend as well for the previous versions of the VictoriaLogs.

dmitryk-dk:
This bug has been fixed in [v0.2.0 release](https://docs.victoriametrics.com/changelog/). Closing the issue as fixed then.

Issue #8, opened 2024-02-26
Labels usage


dmitryk-dk:
If I click on the labels to filter logs via the conditions with the labeks
<img width="1657" alt="Screenshot 2024-02-26 at 12 44 57" src="https://github.com/VictoriaMetrics/victorialogs-datasource/assets/12189570/819b909a-063c-4808-910c-2577f5838b4a">

Datasource enables the filter and shows the correct result. 
<img width="1677" alt="Screenshot 2024-02-26 at 12 45 08" src="https://github.com/VictoriaMetrics/victorialogs-datasource/assets/12189570/a2e7de8e-ecdb-4d08-87fb-57bbc85a26de">

But the response is the same if I remove this filter from the request line and click on the run query.
<img width="1677" alt="Screenshot 2024-02-26 at 12 45 24" src="https://github.com/VictoriaMetrics/victorialogs-datasource/assets/12189570/7dadad17-b096-4fe6-8e40-9625f5f5178f">

It happens because the label stays enabled in the log filter. 
<img width="764" alt="Screenshot 2024-02-26 at 12 56 44" src="https://github.com/VictoriaMetrics/victorialogs-datasource/assets/12189570/d5027b52-b880-4ae6-88e0-7e1b86af2a24">

dmitryk-dk:
This bug has been fixed in [v0.2.0 release](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.2.0). Closing the issue as fixed then.

PR #13, opened 2024-03-10
Bug fixes for query expression and entries limit checks

Loori-R:
- Fixed a bug where a manually removed filter would persist in the query after the "Run query" button was clicked. (#8)
- Fixed query handling to correctly apply the `_time` filter across all queries. (#12)
- Fixed an issue where sometimes an empty response was returned despite having data in VictoriaLogs. (#10)
- Added a client-side record limit check for VictoriaLogs `< v0.5.0` support. (#9)


dmitryk-dk (approved):
LGTM!

Issue #22, opened 2024-05-16
Support vlogs new features: pipes and stats (pipe confuses the parser)


grinapo:
Recent victorialogs supports marvelous features, namely **pipes**. However it breaks the logic in datasource:

`cannot parse query [_time:[2024-05-16T11:56:06.733Z, 2024-05-16T17:56:06.733Z] AND (severity:warning | stats count() logs)]: unexpected token "|" instead of ')'; context: [ 2024-05-16T17:56:06.733Z] AND (severity:warning |]`

The problem is that time gets inserted as 
`_time:XX   AND ( <query_string> )` 
and pipe shall be at the end, outside of the parens. The easiest is probably simply put everything after and including the first pipe outside the parentheses.

hagen1778:
@Loori-R shall we remove support of `_time` filter in favour of Grafana's time picker?

Loori-R:
start and end query args The `_time` value is inserted from Grafana's time picker. We need to replace `_time` with `start` and `end` query args, similar to what was done in [PR #6296](https://github.com/VictoriaMetrics/VictoriaMetrics/pull/6296).

Loori-R:
@grinapo, As a temporary solution, you can manually specify `_time` in the expression. In this case, it will not be set automatically, allowing you to use **pipes** until we fix this issue.

dmitryk-dk:
Hi @grinapo ! The latest [release](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.2.2) should fix your issue. Could you check it?

grinapo:
@dmitryk-dk as far as I am able to test it works flawlessly. Thank you!

dmitryk-dk:
Close as solved. Thank you!

PR #27, opened 2024-06-18
fix: change time range to start and end query args

Loori-R:
change time range limitation from `_time` in the expression to `start` and `end` query args.

#22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## tip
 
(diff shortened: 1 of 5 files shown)


hagen1778 (approved):
LGTM

hagen1778:
Hm, tests are failing https://github.com/VictoriaMetrics/victorialogs-datasource/actions/runs/9566645168/job/26471202226?pr=27

Is this related?

dmitryk-dk:
> Hm, tests are failing https://github.com/VictoriaMetrics/victorialogs-datasource/actions/runs/9566645168/job/26471202226?pr=27
> 
> Is this related?

Yeah, @Loori-R added additional parameters, so that the query URL changed. I think @Loori-R can fix those test before merge

Issue #23, opened 2024-05-20
Error when using `fields` pipe


zekker6:
To reproduce try using the following query: `_time:10h * | unpack_json from _msg | fields level`

The following error is displayed in UI when using the `fields` selector:
> Query error
> An error occurred within the plugin

Grafana logs:
```
grafana logger=context userId=4 orgId=1 uname=z.bessarab@victoriametrics.com t=2024-05-20T14:59:34.689886559Z level=error msg="Internal server error" error="[plugin.downstreamError] client: failed to query data: Failed to query data: rpc error: code = Unknown desc = frame has different field lengths, field 0 is len 0 b
ut field 2 is len 1000" remote_addr=95.217.16.43 traceID=
grafana logger=context userId=4 orgId=1 uname=z.bessarab@victoriametrics.com t=2024-05-20T14:59:34.690327282Z level=error msg="Request Completed" method=POST path=/api/ds/query status=500 remote_addr=95.217.16.43 time_ms=56 duration=56.145455ms size=116 referer="https://grafana.sandbox.victoriametrics.com/explore?orgId
=1&panes=%7B%229y4%22%3A%7B%22datasource%22%3A%22fdmacvu8gcwlce%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22datasource%22%3A%7B%22type%22%3A%22victorialogs-datasource%22%2C%22uid%22%3A%22fdmacvu8gcwlce%22%7D%2C%22expr%22%3A%22_time%3A10h+%2A+%7C+unpack_json+from+_msg+%7C+fields+level%22%2C%22queryType%22%3A%2
2range%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&schemaVersion=1" handler=/api/ds/query status_source=downstream
```

Loori-R:
@dk, can you take a look? 
I think the error occurs because the logs are missing `_time`, `_msg`, and `_stream`, and the entry looks something like `{"level":""}`.

dk:
Sure thing. Looks like the problem is EWRONGDK 🤣

Loori-R:
Oops, sorry :sweat_smile:. I meant @dmitryk-dk

dmitryk-dk:
Yeah, i will check it. What to fix some bugs for the datasource

dmitryk-dk:
[...]

PR #29, opened 2024-06-21
improve response from backend when there are different fields returns

dmitryk-dk:
Fixed issue if the response returns different fields. It creates more generic behavior to prepare responses for UI. 
Sometimes, the response can't include a time value and line values, so it should be generated manually from the returned label values

Related issue: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## tip
 
+* BUGFIX: fix issue when VictoriaLogs returns different fields without stream labels. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/23). 
+
 ## v0.2.1
 
 * BUGFIX: change the `metrics` flag from `false` to `true` in `plugin.json` to ensure the plugin appears in the Grafana datasource selection list.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## tip
 
-* BUGFIX: fix issue when VictoriaLogs returns different fields without stream labels. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/23). 
+* BUGFIX: fix bug with displaying responses with a custom set of fields. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/23).
 
 ## v0.2.1
 
--- a/pkg/plugin/response.go
+++ b/pkg/plugin/response.go
@@ -86,6 +86,9 @@ func parseStreamResponse(reader io.Reader) backend.DataResponse {
 		}
 		labelsField.Append(d)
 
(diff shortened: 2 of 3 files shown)


pkg/plugin/response.go
```diff
@@ -84,6 +85,16 @@ func parseStreamResponse(reader io.Reader) backend.DataResponse {
...
 		labelsField.Append(d)
+
+		if timeFd.Len() == 0 {
+			lineField.Append(string(d))
+		}
+	}
+
+	if timeFd.Len() == 0 {
+		for i := 0; i < lineField.Len(); i++ {
+			timeFd.Append(time.Now())
```
hagen1778, line 96: timestamp should be calculated once. Otherwise, lines could have different timestamps.

Another question: why this check happens here and not at L:89?

hagen1778 (changes requested):
Few comments. And please add a changelog explaining this change.

pkg/plugin/response.go
dmitryk-dk: The frame should have the same number of elements for each field. If one of the fields has a different number of values, the plugin shows the error because we should correctly show the lines per each request.
For example, if timeFd has ten values and labelsField has nine values, we will have an error saying that the number of values can't differ for each frame. The frame is representing one record in the UI

hagen1778: > For example, if timeFd has ten values and labelsField has nine values, we will have an error saying that the number of values can't differ for each frame

Do we have a test case for this?

pkg/plugin/response.go
```diff
@@ -84,6 +85,19 @@ func parseStreamResponse(reader io.Reader) backend.DataResponse {
 			return newResponseError(err, backend.StatusInternal)
 		}
 		labelsField.Append(d)
+
+		// some logsql requests can return only labels or fields without _time field
+		// in that case we need to fill time field with current time value
+		// to avoid empty time field in the response. Please see the test for more details.
```
hagen1778, line 91:
```suggestion
		// Grafana expects timeFd to be always non-empty.
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## tip
 
+* BUGFIX: fix issue when VictoriaLogs returns different fields without stream labels. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/23).
```
hagen1778, line 5:
```suggestion
* BUGFIX: fix bug with displaying responses with custom set of fields. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/23). 
```

pkg/plugin/response.go
dmitryk-dk: fixed

hagen1778 (approved):

Issue #45, opened 2024-07-10
Backend error: frame has different field lengths


clrxbl:
Slack convo: https://victoriametrics.slack.com/archives/CGZF1H6L9/p1720523316896539?thread_ts=1720475100.442039&cid=CGZF1H6L9

I'm trying to visualize a query using a timeseries panel. The following query works to get me the following output:
![image](https://github.com/VictoriaMetrics/victorialogs-datasource/assets/28601081/e4cbd3c0-8340-4387-9456-d2989b875bb8)
```
kubernetes_labels.app.kubernetes.io/name:matchmaker AND kubernetes_namespace_name:$namespace AND "Match found" | stats count() AS count
```

But to visualize it in Grafana I need a query value that is just a number so I add `| format '<count>'` to the query. This breaks things however and returns the following error

```
level=error msg="Internal server error" error="[plugin.downstreamError] client: failed to query data: Failed to query data: rpc error: code = Unknown desc = frame has different field lengths, field 0 is len 2 but field 2 is len 1"
```

dmitryk-dk:
Hi @clrxbl ! Can you share the response from the VictoriaLogs if you execute this query via API?

clrxbl:
> Hi @clrxbl ! Can you share the response from the VictoriaLogs if you execute this query via API?

```
{"count":"507","_msg":"507"}
```
vmui displays it properly aswell. Grafana however, with the exact same query

![CleanShot 2024-07-11 at 15 24 53@2x](https://github.com/VictoriaMetrics/victorialogs-datasource/assets/28601081/6b2aa898-4b6f-4639-96a0-d94028ac0e13)

dmitryk-dk:
> {"count":"507","_msg":"507"}

Yeah, I found the reason I added tests and will create a PR for that fix.
Thank you for your information. Need to test my solution

dmitryk-dk:
@clrxbl ! I have created the request. Here is the response from the explorer.
Would it work for you?
<img width="1791" alt="Screenshot 2024-07-11 at 15 40 27" src="https://github.com/VictoriaMetrics/victorialogs-datasource/assets/12189570/1ca9ed31-56f7-4ca3-9874-cfa619658044">

dmitryk-dk:
@clrxbl i have created a PR you can build it from the my latest commit and check

clrxbl:
> @clrxbl i have created a PR you can build it from the my latest commit and check

[...]

PR #46, opened 2024-07-11
fix double line response to show only numbers from the response

dmitryk-dk:
I removed double-adding lines from the data source response. It produced a problem when the user tries to count some values.

Related issue: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/45


Loori-R (approved):
LGTM!

Issue #48, opened 2024-07-18
Support Builder Mode in Grafana 


tenmozes:
We need to add support for autocompletion for labels in Builder mode in Grafana 

the same feature in Loki 

![builder-ac](https://github.com/user-attachments/assets/8d436704-e3e2-41bd-bbb5-8db810c6a115)

Loori-R:
**Roadmap for Future Improvements:**

- [ ] **Support for the NOT operator**. Example expressions: `_time:5m error NOT buggy_app` and `_time:5m log.level:error !(buggy_app OR foobar)`.
- [ ] **Support for creating nested queries directly from the UI builder**. Currently, it can parse nested queries but does not allow users to create them.
- [ ] **Support for using pipes in queries**. The list of available pipes can be found [here](https://docs.victoriametrics.com/victorialogs/logsql/#filters).
- [ ] **Support for various LogsQL filters**. The complete list of filters is available [here](https://docs.victoriametrics.com/victorialogs/logsql/#filters).
- [ ] **Support for expressions where application identifiers (e.g., `buggy_app` and `foobar`) are stored in a specific field**. For example: `_time:5m log.level:error !(app:buggy_app OR app:foobar)`

archef2000:
Is there a plan for when pipes (more specifically the format pipe) will be implemented? Is a PR for this feature welcome? I come from loki, but like the ease of HA in VM and was already using it for metrics so I also switched to VL.

Loori-R:
Hi, @archef2000! Pipe support is planned, but not in the near future. We’d be happy to review a PR if someone is up for it!

archef2000:
First as Queries are also supported after a pipe I would combine filters and operations and just select a operation like ["label filter", "matcher", etc].

Second I have some problems testing my changes. What is your workflow for testing frontend changes?

Loori-R:
1. I agree, but we need to be careful here — some pipe operations allow parts of the phrase to be omitted. For example, `sort` can be written as `sort by`, `| sort`, or `| order by`, so we need to handle those variations properly.

2. For testing, we have test files like `src/components/QueryEditor/QueryBuilder/utils/parsing.test.ts`. You can add your own tests there if needed. If necessary, feel free to add your own test files as well.

archef2000:
No I mean how are you quickly testing UI changes in Grafana, as It tkes me a lot of time building the ui everytime I change something and restart grafana.

archef2000:
[...]

PR #57, opened 2024-08-13
add beta version of the query builder

Loori-R:
Added a UI for the query builder.
  - The builder allows creating simple queries based on field names and field values.
  - The list of loaded values is cached for improved performance.
  - The builder can parse nested queries, such as:
    ```
    (foo and foo:bar) or ((foo or bar) and (foo and (foo or bar)))
    ```

See the list of planned improvements in [issue #48](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/48).
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## tip
 
+* FEATURE: add beta version of the query builder. The builder allows selecting `field names` and `field value`. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/48)
+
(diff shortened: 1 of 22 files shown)


dmitryk-dk (approved):
LGTM!

hagen1778:
Hi @Loori-R!
The builder works for me, but it becomes very laggy when selected field has many options to select. 
For example, selecting _msg field staggers my browser for couple of seconds 
image

I think that we should limit the number of options we load into interface.

hagen1778 (approved):

Issue #65, opened 2024-08-22
incorrect handling of variables with special characters and multiple values


Loori-R:
Hi! Thank you for feature. I tried this out and found some bugs.

Multi-value seems doesn't work. I have query: 
`consumer: $consumer`
Choose 2 options and get nothing. From google devtools in network i see following query:
`consumer: <null>|admin`
I think it should have looked like this:
`consumer: ("<null>" OR "admin")`

Query works badly with values with special characters. I have query:
`proxy_list: $proxy_list`
From google devtools in network i see following query:
`proxy_list: api_name:clientid-persons-v1;v1\\.14\\.1`
I think it should have looked like this:
`proxy_list: "api_name:clientid-persons-v1;v1.14.1"`

I think need to add quotes and work through the moment with multi-value

_Originally posted by @Neko-Follower in https://github.com/VictoriaMetrics/victorialogs-datasource/issues/40#issuecomment-2304239395_

pgassmann:
multi value doesn't work yet as expected. 
I would like to filter logs based on multiple selected values, of e.g. the `host` label

if using regexp filter `host:~"$host"` it partially works, but matches anything with thatsubstring. also, [regexp filter](https://docs.victoriametrics.com/victorialogs/logsql/index.html#regexp-filter) is not good for performance

ideally the value could be passed to the [multi exact `:in()` filter](https://docs.victoriametrics.com/victorialogs/logsql/index.html#multi-exact-filter).

`host:in($host) and compose_project:~"$compose_project" and $log_query`

![grafik](https://github.com/user-attachments/assets/3d043142-7a9b-46f6-a249-8bb5881ab9b6)

Workaround:
regexp filter with start `^` and end `$` matcher in single quotes:
`host:~'^$host$' and compose_project:~'^$compose_project$' and $log_query`

also works in the variables section for filtering field_value results with query
![grafik](https://github.com/user-attachments/assets/560ca016-0d15-4ee1-a379-2b90d32fd4be)
![grafik](https://github.com/user-attachments/assets/a490d55e-720b-48e6-9402-944fa7a97141)

dmitryk-dk:
Hi @Loori-R ! The issue was fixed in the [v0.4.0](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.4.0) release. Fill free to reopen the issue if you find some problems

Issue #64, opened 2024-08-22
Dynamic variables using label values got: "Error updating options: this is undefined"


2333:
Hi, thanks for updating datasource with dynamic variables, I'm really looking forward to use it on my dashboard. So when I updated my plugin when I see it's released.
But when I'm trying to use it, I've got an error: "Error updating options: this is undefined", any hinds of this?
My grafana version is v10.1.2 and victorialogs-datasource is v0.3.0.
![image](https://github.com/user-attachments/assets/22904762-faf1-4415-bfe7-120a6bd16a9f)
thank you.
BR

Loori-R:
Hi @2333, could you please share the error from the DevTools console?

2333:
> Hi @2333, could you please share the error from the DevTools console?

Hi @Loori-R, here's console output, plz let me know if you need more information.

![image](https://github.com/user-attachments/assets/6ad3153c-96dd-429f-960d-0c8ee365788e)

dmitryk-dk:
Hi @2333 ! The issue was fixed in the [v0.4.0](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.4.0) release. Fill free to reopen the issue if you find some problems

PR #66, opened 2024-08-27
fix: handle dynamic variables

Loori-R:
- fix dynamic variable issue causing `this is undefined` error. 
   Ref #64.
- fix multi-value variable handling to properly format queries. 
   Ref #65.
- fix issue with special characters in variable values. 
   Ref #65.
- add tests to cover these fixes and ensure correct functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## tip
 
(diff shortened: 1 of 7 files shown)


hagen1778 (approved):
LGTM

PR #69, opened 2024-09-04
add replace  value

### LINKED ISSUE #61: More panels support for plugin
That would be awesome to add support for the following panel types
- stat
- pie 
- time series
- gauge
- bar

yincongcyincong:
fix #61
support $__interval variable in query

### DIFF
--- a/pkg/plugin/query.go
+++ b/pkg/plugin/query.go
@@ -6,6 +6,8 @@ import (
 	"path"
 	"strconv"
 	"time"
+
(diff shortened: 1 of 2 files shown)


dmitryk-dk (approved):
LGTM! @yincongcyincong Thank you for the pull request

pgassmann:
Thank you for supporting the $__interval variable I requested.

Note: there are more global variables in Grafana that might be useful in queries like $__range that users of prometheus or other datasources might expect. https://grafana.com/docs/grafana/latest/dashboards/variables/add-template-variables/#global-variables

Issue #72, opened 2024-09-06
Query filter does not work


Neko-Follower:
Hi. I want to use query filter, but it doesn't seem to work
![image](https://github.com/user-attachments/assets/6563c1c6-d1ee-459b-84a2-60ea2c2aca89)

dmitryk-dk:
Hi @Neko-Follower ! I have checked this question on my local, and it works. See examples below
1. With all values
<img width="1799" alt="Screenshot 2024-09-06 at 10 48 46" src="https://github.com/user-attachments/assets/d6e894db-2917-4921-99de-045806700032">

2. With only stdout value
<img width="1799" alt="Screenshot 2024-09-06 at 10 49 50" src="https://github.com/user-attachments/assets/fff49fc9-0559-454e-8986-c215fc22c5ab">

Could you please share the version of the grafana?

Neko-Follower:
I removed the space. It didn't work
![Снимок экрана 2024-09-06 в 15 41 06](https://github.com/user-attachments/assets/c6a1619e-cde5-4103-9d37-59ff2676633a)

My Grafana version - 11.1.4
I saw in #71 issue you use Grafana version 9.5.21. How you use it if datasource requires version >= 10.0.3?
![Снимок экрана 2024-09-06 в 15 46 03](https://github.com/user-attachments/assets/ef5d3c6e-8432-4cf6-bccb-0295ebe68871)

I'll try to downgrade Grafana to 10 and test.

dmitryk-dk:
Checked with 11.1.4. It takes some time to return a response, and maybe you have a proxy in between the Grafana and Victorialogs that may stop the request for some time. It is just my assumption because you have the error like 504 gateway timeout. But can you check that you do not have any proxy
<img width="1800" alt="Screenshot 2024-09-06 at 13 03 31" src="https://github.com/user-attachments/assets/9ecacc75-998c-4b86-9893-9e24f81f5070">

Neko-Follower:
You are right. It is error 504 gateway timeout. I looked and there is no proxy between grafana and victorialogs.
Also I looked at victorialogs logs:
```
[...]

Issue #71, opened 2024-09-06
Timeout Error when querying variables


Neko-Follower:
Hi! I run into some bugs with new v0.4.0 version. I try to create variable namespace and get 504 Timeout error
![Снимок экрана 2024-09-06 в 12 24 56](https://github.com/user-attachments/assets/0bf79034-c6cd-48c6-ac07-ac0078428fdc)
![Снимок экрана 2024-09-06 в 12 30 46](https://github.com/user-attachments/assets/352d11f2-10a7-4b8f-8c66-ca9650fd5c89)
![Снимок экрана 2024-09-06 в 12 32 13](https://github.com/user-attachments/assets/8413100b-ac04-43bb-a10c-327daa48e271)

For testing purposes I switch back to v0.3.0 version. And I didn't find any errors here.
![Снимок экрана 2024-09-06 в 12 35 04](https://github.com/user-attachments/assets/9d37dc80-2b7d-4e98-aeec-fbd322a82bc0)

dmitryk-dk:
Hi @Neko-Follower ! Is it reproduced with the previous version?

dmitryk-dk:
@Neko-Follower ! I have checked the behavior and it works well with my local installation
<img width="1346" alt="Screenshot 2024-09-06 at 10 40 55" src="https://github.com/user-attachments/assets/ea8ca93f-09cd-4c88-a741-264216c3d400">

I used grafana 9.5.21, and the v0.4.0 of the datasource

dmitryk-dk:
It takes some time to query those values. Can you check that you do not have any proxy between Datasource and Victorialogs?

dmitryk-dk:
Hi @Neko-Follower ! We have discussed all described issues. I need to do some testing on both issues and will be back soon. I think I will make all the checks today. But I need to prepare for some of the tests at my local env.

Neko-Follower:
Hi @dmitryk-dk ! Ok, I'll be waiting for results. #72 can be closed since it is related to timeout error

PR #73, opened 2024-09-10
language_provider: fix query time params names

hagen1778:
The correct query param names for /field_values are `start` `end`. See https://docs.victoriametrics.com/victorialogs/querying/#querying-field-values

Sending `from` and `to` params meant time range filter wasn't applied at all. This made VictoriaLogs make the full data scan.
Which probably caused the following issues:
* https://github.com/VictoriaMetrics/victorialogs-datasource/issues/71
* https://github.com/VictoriaMetrics/victorialogs-datasource/issues/72


dmitryk-dk (approved):
LGTM!

Issue #81, opened 2024-10-05
Improve filter logs by button


yuriydzobak:
Hi,
would be nice if i already filter my data and want to filter by button
for example i write LogsQL that have some pipes and functions
```
pod_name:vm-logs-single-0 | unpack_json | rename message as _msg
```
![image](https://github.com/user-attachments/assets/0518534e-2adc-4ce6-af42-87e5a3db2724)
and when you press the button 1. it added in the and AND field:value but if you have function in the end it broke the query
![screenshot-000448](https://github.com/user-attachments/assets/56819a53-e1e1-42e4-bbc3-715ce8739c48)

It works in that logic fine when you have a simple filter by word not the functions. [Example](https://play-grafana.victoriametrics.com/explore?schemaVersion=1&panes=%7B%22vo8%22:%7B%22datasource%22:%22P566C6523BE02F42C%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22victorialogs-datasource%22,%22uid%22:%22P566C6523BE02F42C%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22type:message%20%7C%20user:U06C62ZA6E6%20AND%20display_name:James%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%22now-24h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)
![image](https://github.com/user-attachments/assets/1dbf0150-621c-4639-9a58-15db9fb9d915)
![image](https://github.com/user-attachments/assets/6ff94f9f-e5de-4890-86b9-c3a658caff63)
how to reproduce
![image](https://github.com/user-attachments/assets/5338ef3e-a6c2-49a2-a616-b774a5ff3e88)
![image](https://github.com/user-attachments/assets/5ad0ca6e-9b3e-4b17-be4a-86d089a59f7a)

Loori-R:
The issue should be resolved in the released [v0.6.0](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.6.0) version

PR #90, opened 2024-10-15
fix filter application in queries

Loori-R:
Fixed the application of filtering in queries.
See issue #81 for more details.


dmitryk-dk (approved):
LGTM!

PR #92, opened 2024-10-17
fixed healthcheck

AndrewChubatiuk:
path.Join removes repetitive `/` symbols, which affects protocol prefix, replaced it with less invasive approach


dmitryk-dk (approved):
LGTM!

Issue #91, opened 2024-10-15
bug: VictoriaLogs - missing records (but present in Loki)


arseny-zinchenko:
**Chart name and version**
Helm chart: victoria-logs-single. version: v0.6.3
Docker image: victoriametrics/victoria-logs:v0.35.0-victorialogs
Grafana VictoriaLogs Data source version: 0.5.0

**Describe the bug**
We have a log records with a big `body` field.
They are written to the Loki, but can't be found in VMLogs.

**Loki query**: `{app="backend-api", namespace="prod-backend-api-ns"} |= "b4c2ecd2-000b-46e5-a448-0983963c0688"`
- returns desired records

![image](https://github.com/user-attachments/assets/8e66781d-607d-42aa-a68f-b9f2480bbe28)

**VMlogs query**: `app:="backend-api" namespace:="prod-backend-api-ns" "b4c2ecd2-000b-46e5-a448-0983963c0688"`
- returns noting

![image](https://github.com/user-attachments/assets/05caa98f-3858-46e4-95a0-c0744669078a)

If query by an Amazon X-Ray ID related to the records, then:

**Loki query**: `{app="backend-api", namespace="prod-backend-api-ns"} |= "Root=1-670e6e90-2e398c1005aa39f076ec1428"`
- returns 8 results, including the record with the big `body` field

![image](https://github.com/user-attachments/assets/4e088549-5946-44e5-ab10-d0efc09fde6c)

**VMlogs query**: `app:="backend-api" namespace:="prod-backend-api-ns" "b4c2ecd2-000b-46e5-a448-0983963c0688"`
- returns 7 results, and no record with the big `body` field
![image](https://github.com/user-attachments/assets/9ad4a36e-cc6f-415c-a548-b7bc7957ed2c)

**Custom values**
Please provide only custom values (excluding default ones):
```yaml
  grafana:
    enabled: true
   ...
    env:
      GF_PLUGINS_ALLOW_LOADING_UNSIGNED_PLUGINS: "victorialogs-datasource"
    ...
    plugins:
     ...
      - https://github.com/VictoriaMetrics/victorialogs-datasource/releases/download/v0.5.0/victorialogs-datasource-v0.5.0.zip;victorialogs-datasource
```

AndrewChubatiuk:
transfered issue to victorialogs-datasource as this issue looks like datasource related

dmitryk-dk:
[...]

Issue #80, opened 2024-10-05
Missed data if logs message is huge


yuriydzobak:
Hi,
I'm using plugin version 0.5.0 and victorialogs 0.33.0 i've found an issue in grafana plugin
As you see iam able to see log entries in VMUI
![screenshot-000442](https://github.com/user-attachments/assets/e40e64c6-a09b-4779-abb2-e9fe56c7c911)

If i'm doing the same in grafana, it random show me the data, the limit range is 1k and time range 2 days too, if i select 1 hour limit it has the same problem
![screenshot-000443](https://github.com/user-attachments/assets/99926d3f-dd63-4a63-9cd4-d6dada8c4cf5)

```
_time:2d pod_namespace:monitoring container_name:lf | count()
{"count(*)":"1054"}
```
it has logs 1k entries

If i press button to run query it random show me logs in timeline each time

https://github.com/user-attachments/assets/5144e52b-fed6-4d22-8171-d31d5b91bc08

I beelive the problem im my logs message size
grafana show another logs entries without probelm, but that app provided the message size almost 200kb data in single message
VictroaLogs has message in logs like this
```
{"ts":"2024-10-05T08:38:55.780Z","level":"info","caller":"VictoriaMetrics/app/vlselect/main.go:161","msg":"client has canceled the request after 0.064 seconds: remoteAddr=\"10.10.94.253:59254\", requestURI: \"/select/logsql/query?end=1728117535&limit=1000&query=_time%3A2d+pod_namespace%3Amonitoring+container_name%3Alf+&start=1727944735\""}
{"ts":"2024-10-05T08:38:56.642Z","level":"info","caller":"VictoriaMetrics/app/vlselect/main.go:161","msg":"client has canceled the request after 0.075 seconds: remoteAddr=\"10.10.94.253:59270\", requestURI: \"/select/logsql/query?end=1728117536&limit=1000&query=_time%3A2d+pod_namespace%3Amonitoring+container_name%3Alf+&start=1727944736\""}
{"ts":"2024-10-05T08:38:57.337Z","level":"info","caller":"VictoriaMetrics/app/vlselect/main.go:161","msg":"client has canceled the request after 0.121 seconds: remoteAddr=\"10.10.94.253:59276\", requestURI: \"/select/logsql/query?end=1728117537&limit=1000&query=_time%3A2d+pod_namespace%3Amonitoring+container_name%3Alf+&start=1727944737\""}
[...]

PR #95, opened 2024-10-24
issue-91: allow reading strings with arbitrary length when parsing response in stream mode

dmitryk-dk:
Allow reading strings with arbitrary length when parsing response in stream mode.
Updated tests.

Related issue: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/91, https://github.com/VictoriaMetrics/victorialogs-datasource/issues/80


pkg/plugin/response.go
```diff
@@ -41,10 +42,31 @@ func parseStreamResponse(reader io.Reader) backend.DataResponse {
...
 	lineField.Name = gLineField
 
-	scanner := bufio.NewScanner(reader)
+	sc := bufio.NewReader(reader)
 
-	for scanner.Scan() {
-		value, err := fastjson.ParseBytes(scanner.Bytes())
+	finishedReading := false
+	n := -1
+	for !finishedReading {
```
tenmozes, line 49: regularly it calls next and it's true

pkg/plugin/response.go
```diff
@@ -41,10 +42,31 @@ func parseStreamResponse(reader io.Reader) backend.DataResponse {
...
+			}
+			finishedReading = true
+		}
+
+		if len(b) == 0 {
+			continue
+		}
+
+		b = bytes.Trim(b, "\n")
+		value, err := fastjson.ParseBytes(b)
```
hagen1778, line 69:
```
	// aligns with -insert.maxLineSizeBytes
	br := bufio.NewReaderSize(reader, 256*1024)
	var parser fastjson.Parser
	var finishedReading bool
	for n := 0; !finishedReading; n++ {
		b, err := br.ReadBytes('\n')
		if err != nil {
			if errors.Is(err, bufio.ErrBufferFull) {
				backend.Logger.Info("skipping line number #%d: line too long", n)
				continue
			}
			if errors.Is(err, io.EOF) {
				// b can be != nil when EOF is returned, so we need to process it
				finishedReading = true
			} else {
				return newResponseError(fmt.Errorf("cannot read line in response: %s", err), backend.StatusInternal)
			}
		}

		if len(b) == 0 {
			continue
		}

		value, err := parser.ParseBytes(b)
```

pkg/plugin/response.go
```diff
@@ -41,10 +42,31 @@ func parseStreamResponse(reader io.Reader) backend.DataResponse {
 	lineField := data.NewFieldFromFieldType(data.FieldTypeString, 0)
 	lineField.Name = gLineField
 
-	scanner := bufio.NewScanner(reader)
+	sc := bufio.NewReader(reader)
```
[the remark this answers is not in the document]
dmitryk-dk: I think about this because I checked the logic of the parsing log lines. I decided to use default values because under the hood the bufio.NewReader has an expandable buffer like the one we used in the utils.BytesBuffer. But I will set it :)

pkg/plugin/response.go
dmitryk-dk: updated. Thank you :)

hagen1778 (approved):
LGTM

Issue #83, opened 2024-10-05
Explore show live logs mode


yuriydzobak:
Hi,
As VictoriaLogs ahs [live tailing](https://docs.victoriametrics.com/victorialogs/querying/#live-tailing) mode
It would be nice if plugin will have this feature as well
Loki has it in their plugin
For example open [loki play ](https://play.grafana.org/explore?schemaVersion=1&panes=%7B%2216y%22:%7B%22datasource%22:%22grafanacloud-demoinfra-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bcluster%3D%5C%22do-nyc1-demo-infra%5C%22%7D%20%7C%3D%20%60%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-demoinfra-logs%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)

![screenshot-000456](https://github.com/user-attachments/assets/8906485b-e542-4a0a-ad89-1442cec14d42)
and it expand ui in explore mode 
![screenshot-000457](https://github.com/user-attachments/assets/796df229-7330-4622-9c15-0d151dee529d)
Maybe Loki has a refresh mode instead of a live tailing mode like VictoriaLogs. However, it would be a useful feature for debugging apps in live mode.

dmitryk-dk:
Hi @yuriydzobak ! The feature was added in the [release](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.7.0). Could you please check it? I close the issue. If you find any problems with new release please reopen the issue

maxpain:
@dmitryk-dk I've just checked, and it works perfectly!

maxpain:
I think I've found the bug. The "Clear logs" button breaks getting the new log entries.

maxpain:
![image_2024-11-07_16-15-31 (1)](https://github.com/user-attachments/assets/fa66e1cb-4029-400a-83a4-7857b69d1784)

maxpain:
I also found a strange bug. The live mode ignores my filter for some reason.

https://github.com/user-attachments/assets/88a0be01-c2d5-43fc-814c-00f262db05ce

dmitryk-dk:
[...]

PR #101, opened 2024-10-31
issue-83: enable live tail requests

dmitryk-dk:
Support for the live display logs in the Explore panel was added. It uses a tail endpoint to make requests. 

new_feature

Related issue: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/83


Loori-R (approved):
LGTM!

PR #104, opened 2024-11-13
update dev process

dmitryk-dk:
Starting from Grafana v10.4.0 it tries to check the folder inside the defined plugin folder.
If we set 
`plugins = /Users/path/victorialogs-datasource` Grafana tries to check source plugins in the `src` folder and shows the error that the backend file can be found
But if we declare the path like
`plugins = /Users/path/plugins/victorialogs-datasource` (we create a folder above our plugin) it works correctly. 

Also started from Grafana v10.4.0 when we started it and defined the plugin folder it added its own plugin named
`grafana-lokiexplore-app`

Those changes are compatible with the previous version of Grafana


Loori-R (approved):
LGTM!

PR #106, opened 2024-11-14
Issue 61/more panels support

dmitryk-dk:


Loori-R (approved):
LGTM!

src/datasource.ts
```diff
@@ -294,3 +299,13 @@ export class VictoriaLogsDatasource
...
 }
+
+function determineQueryType(panelPluginId?: string) {
+  if (panelPluginId && !['logs', 'table', 'timeseries'].includes(panelPluginId)) {
+    return QueryType.Stats;
+  }
+  if (panelPluginId === 'timeseries') {
+    return QueryType.StatsRange;
+  }
+  return QueryType.Instant;
```
hagen1778, line 310: Can this be a swtich statement?

pkg/plugin/query.go
```diff
@@ -4,27 +4,52 @@ import (
...
+	RefID           string `json:"refId"`
+	Expr            string `json:"expr"`
+	LegendFormat    string `json:"legendFormat"`
+	MaxLines        int    `json:"maxLines"`
+	TimeRange       TimeRange
+	IntervalMs      int64  `json:"intervalMs"`
+	Interval        string `json:"interval"`
+	TimeInterval    string `json:"timeInterval"`
+	StatsQueryRange bool   `json:"statsQueryRange"`
+	StatsQuery      bool   `json:"statsQuery"`
```
hagen1778, line 49: What are these two fields? Do you use them?

pkg/plugin/query.go
```diff
@@ -4,27 +4,52 @@ import (
...
 // Query represents backend query object
 type Query struct {
-	RefID        string `json:"refId"`
-	Expr         string `json:"expr"`
-	LegendFormat string `json:"legendFormat"`
-	MaxLines     int    `json:"maxLines"`
-	TimeRange    TimeRange
-	IntervalMs   int `json:"intervalMs"`
-	url          *url.URL
+	RefID           string `json:"refId"`
```
hagen1778, line 40: Seems like most of the fields here are just `backend.DataQuery` fields. Should we embedd this struct instead?

pkg/plugin/datasource.go
```diff
@@ -228,7 +228,8 @@ func (d *Datasource) query(ctx context.Context, _ backend.PluginContext, query b
 	}
 
 	q.TimeRange = TimeRange(query.TimeRange)
-
+	q.MaxDataPoints = query.MaxDataPoints
```
hagen1778, line 231: Would it be better to parse `query backend.DataQuery`  before this function call on L:131? It would be more clear and testable, I think.

pkg/plugin/response.go
```diff
@@ -277,3 +298,132 @@ func labelsToJSON(labels data.Labels) (json.RawMessage, error) {
...
+	Result     json.RawMessage `json:"result"`
+}
+
+// Response contains fields from query response
+type Response struct {
+	Status string `json:"status"`
+	Data   Data   `json:"data"`
+}
+
+type promInstant struct {
```
hagen1778, line 333: `promInstant` has identical structure as `promRange`. Are you sure these need to be two different structs? Should we change the name of the struct to remove `prom` connotation as it seems irrelevant here and rather is a result of copypaste

pkg/plugin/response.go
```diff
@@ -277,3 +298,132 @@ func labelsToJSON(labels data.Labels) (json.RawMessage, error) {
...
+
+		frames[i] = data.NewFrame("",
+			data.NewField(data.TimeSeriesTimeFieldName, nil, timestamps),
+			data.NewField(data.TimeSeriesValueFieldName, data.Labels(res.Labels), values))
+	}
+
+	return frames, nil
+}
+
+type promScalar Value
```
hagen1778, line 389: vlogs doesn't return scalar types in stats response - see here https://github.com/VictoriaMetrics/VictoriaMetrics/tree/master/app/vlselect/logsql

hagen1778 (changes requested):
Thanks! Please see my comments

I didn't have chance to test it locally. Will try to do it once I have a chance.

pkg/plugin/query.go
dmitryk-dk: I'm sorry. It's my fault. I forgot to remove them. The first implementation was with those fields after we decided to use QueryType.

pkg/plugin/query.go
dmitryk-dk: fixed

pkg/plugin/datasource.go
dmitryk-dk: yes. Made this change

pkg/plugin/response.go
dmitryk-dk: used one struct and simplify a little `getDataFrames` method

pkg/plugin/response.go
dmitryk-dk: removed it.

dmitryk-dk:
> Thanks! Please see my comments
> 
> I didn't have chance to test it locally. Will try to do it once I have a chance.

Thank you for your comments! Always simplified code :). I made changes and checked, works well. 
But please check from your side, maybe you will find some incorrect behavior.

Issue #119, opened 2024-11-23
Misleading stat output for all time instead of time range in v0.8.0


p5i:
When using the 'Stat', 'Gauge' or other similar visualizations in Grafana, the stats query on the dashboard
`* | stats count()`
transforms to this query to VL
`POST /select/logsql/stats_query?query=%2A+%7C+stats+count%28%29&time=1732387652` 
There is only one time constraint. Because of this, instead of getting stats for the specified on the dashboard time range, we receive stats for all the time before the upper bound.
![photo_1](https://github.com/user-attachments/assets/f8579072-0a54-43b5-be63-6a61f933e601)

This looks misleading because for Grafana it is common to receive data only for the time period specified on the dashboard.

This can be worked around by specifying the time filter with Grafana's global variables '$__from' and '$__to'.
`_time: [$__from, $__to] | stats count()` 
![photo_2](https://github.com/user-attachments/assets/b2fed7a1-d5bc-4712-97e9-54be3726387a)

But it looks hacky for such a common query.

I think it would be better to transform the stats query in Grafana to the VL query with time range
```
$ curl -s http://localhost:9428/select/logsql/stats_query -d 'query= * | stats count()' -d 'start=now-1d' -d 'end=now' | jq
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {
          "__name__": "count(*)"
        },
        "value": [
          1732388516.0014431,
          "2167582"
        ]
      }
    ]
  }
}
```
This form of stats_query is not documented here [https://docs.victoriametrics.com/victorialogs/querying/#querying-log-stats](https://docs.victoriametrics.com/victorialogs/querying/#querying-log-stats) , but I think it should be. And it works correctly as far as I can tell.

p5i:
Hi @dmitryk-dk. Thanks for the answer. I think  #111 is exactly what I asked here and https://github.com/VictoriaMetrics/victorialogs-datasource/issues/112 can help to optimize the queries with the time filter.

hagen1778:
@dmitryk-dk @Loori-R I think what @p5i says about instant panels has a lot of sense.

It is unlikely someone will need to have a panel on dashboard that does `stats` over the all retentionPeriod, as it is very expensive calculation. I think it would be reasonable to do the following changes to stats panels:
* automatically apply time filter to all instant panels. I.e. `* | stats count()` should be automatically transformed by backend to `* _time:$__range | stats count()`
[...]

Issue #112, opened 2024-11-20
Support $__range variable in queries


Loori-R:
Add support for the $__range variable in queries. This would enable queries like `_time: $__range | stats by(host) count()` to retrieve stats for the time range selected in the Grafana time picker.

_Originally posted by @hagen1778 in https://github.com/VictoriaMetrics/victorialogs-datasource/issues/61#issuecomment-2487919910_

dmitryk-dk:
Hi @Loori-R ! The issue was fixed in the new [release](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.9.0). Close as implemented. Feel free to reopen if you find any problems.

PR #121, opened 2024-11-25
add support for the `$__range` variable

dmitryk-dk:
Added support for the `$__range` variable. It will be transformed to the `[time_from, time_to]` in the Unix format. 
For example Grafana from: `now-1h`, to: `now` will be transformed `[1732546325, 1732549925]`. 
It is necessary to use `_time:` token with this Grafana variable to build correct requests.

Query example:
1.  `_time:$__range | stats count()` will be transformed to `_time:[1732320000, 1732492800] | stats count()`
2. `* _time:$__range` will be transformed to `* _time:[1732320000, 1732492800] | stats count()`

Related issues: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/112, https://github.com/VictoriaMetrics/victorialogs-datasource/issues/119

Variable `$__from, $__to` works as is, so we do not need to add support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## tip
 
+* FEATURE: Add support for the `$__range` variable in queries.  It will be transformed to the `[time_from, time_to]` in the Unix format. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/112).
+
 ## v0.8.0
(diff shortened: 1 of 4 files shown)


hagen1778 (dismissed):
LGTM! But missing a changelog note

Loori-R (approved):

Issue #118, opened 2024-11-22
Time range resets to 24 hours on variables refresh in Grafana 11.3 when using query variables


p5i:
After updating Grafana to v11.3.0 or v11.3.1 time range resets to 24 hours when variables are refreshed.
To reproduce you can add variable of type 'query', make it refresh on time range change, than build the time series graph.
Try to change time range to less then 24 hours and it will remain 24 hours.

![photo](https://github.com/user-attachments/assets/859a5d9e-17cf-4d03-ae01-c0e0d204a515)

I tried the datasource versions v0.8.0 and v0.7.0

koriit-kontakt:
Hi, we are having the same issue. It started happening after upgrading to Grafana v11.3.0 from v11.2.2. We tested it with datasource versions v0.6.1 and v0.9.0.

@p5i the workaround of not having variables refresh on time range change is also working in our case, thanks a lot!

dmitryk-dk:
Hi @p5i , @koriit-kontakt ! I Suppose this issue is related to this [issue](https://github.com/grafana/grafana/issues/95443)

p5i:
Hi @dmitryk-dk 
Thanks for the answer.
That issue may be related, but not exactly the same, according to the description.
I don't have InfluxDB to check if there is similar problem.
I use also Graphite and Loki datasources and there is no such issues.

My problem is that time range of the query differs from time range specified on the dashboard. For example I set 1 munute range on the dashboard, but in the query I see 24 hours (23:59:59 exactly) range.

On the dashboard:
![2024-12-05 12-12-29](https://github.com/user-attachments/assets/e1068c18-431c-4356-ba41-aa823d4f88d2)

In the Query Inspector:
![2024-12-05 12-12-55](https://github.com/user-attachments/assets/a58a15e3-97df-402f-a96b-bf1dd790f63d)

The query to VL from the network dump:
```POST /select/logsql/stats_query_range?end=1731963599&query=%2A%7C+stats+count%28%29&start=1731877200&step=1m0s HTTP/1.1 ```

```
date -u -d @1731877200
Sun Nov 17 21:00:00 UTC 2024

date -u -d @1731963599
Mon Nov 18 20:59:59 UTC 2024
```



Reproduced with Grafana v11.3.2 and victorialogs-datasource v0.10.0

hagen1778:
[...]

PR #137, opened 2024-12-06
fix time range reset with query variables

Loori-R:
This PR fixes a bug where the time range reset when using query variables.  

Cause:
The issue was caused by using `startOf` and `endOf` on `TimeRange` objects, which led to mutations.

Related Issue: #118

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## tip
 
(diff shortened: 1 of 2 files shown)


dmitryk-dk (approved):
LGTM!

hagen1778 (approved):
Thanks!

PR #140, opened 2024-12-06
makefile cleanup

AndrewChubatiuk:
same as https://github.com/VictoriaMetrics/victoriametrics-datasource/pull/213


dmitryk-dk (commented):
Hi! I just want to merge  the PR but tested it and found some inconsistency.
It shouldn't build on different folders. It should be in one folder. Also it doesn't contain the tar archive. 
We will change the plugin id so i need to change make files and other staff in the repository. Those changes will produce the conflicts with your PR. I will going to reuse some logic from your PR if you don't mind.
Screenshot 2024-12-16 at 12 58 51

AndrewChubatiuk:
@dmitryk-dk 
why do we need tar archive?

dmitryk-dk:
> @dmitryk-dk why do we need tar archive? all results are in dist folder or what do you mean?

> all results are in dist folder or what do you mean?

FE and backend files should be collected in one folder. Named as plugin id.
The archive should just contain that folder. Because Grafana, by default, relies on folder not just on files.

> why do we need tar archive?

When we started to create the first plugin for VictoriaMetrics we decided to use the same approach as VictoriaMetrics has. When we added support for Windows we added zip archive. In that case we need both types of archive

AndrewChubatiuk:
@dmitryk-dk  added tar and plugin id dir to a artifact root

dmitryk-dk (approved):
LGTM!

dmitryk-dk (approved):
LGTM!

Loori-R (approved):

Issue #142, opened 2024-12-06
Add time field if it is not present for sats API call


dmitryk-dk:
Stats API doesn't support the time ranges for requests. To make it possible we should add `_time` field automatically if user didn't set it

PR #143, opened 2024-12-06
issue-142: add `_time` field if it doesn't present in the query

dmitryk-dk:
Automatically add `_time` field before the main query or the first pipe (because we can use _time in the stats by function) for the stats API call. 

Related issue: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/142


Loori-R (approved):
LGTM!

Loori-R (approved):

PR #151, opened 2024-12-10
fix the check for the stats pipe

Loori-R:
Fix the check for stats pipe functions in expressions.  
This resolves unnecessary warnings for `Range` and `Instant` queries.


hagen1778 (approved):

dmitryk-dk (approved):
LGTM!

PR #161, opened 2024-12-16
issue-156: change plugin id to Grafana naming convention

dmitryk-dk:
Renamed plugin id to Grafana naming convention. Updated README.md with related changes.

I checked how it works and how the Grafana checks new plugin name
Screenshot 2024-12-16 at 18 12 04

To run those changes it is required to update the defaults.ini  or grafana.ini file with the new plugin name in the `allow_loading_unsigned_plugins` field.
```
allow_loading_unsigned_plugins = victoriametrics-logs-datasource
```

If you have provision files - it is required to update `type` field to the `victoriametrics-logs-datasource` plugin name

Related issue: #156

PR #167, opened 2024-12-17
Fix wrong plugin download urls

risen228:
It looks like the README contains incorrect URLs for downloading the plugin from the release assets. I discovered this when I got a 404 when installing.

Here is the change that broke URLs: https://github.com/VictoriaMetrics/victorialogs-datasource/pull/161

I don't know what the plans are, maybe the file name will be updated in the next release, but it's not like that now 🤔

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Please find the example of provisioning Grafana instance with VictoriaLogs datas
        grafana:
          image: grafana/grafana:11.0.0
          environment:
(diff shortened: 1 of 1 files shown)


AndrewChubatiuk:
@risen228 plugin url was changed intentionally, please take a look at a changelog

risen228:
Ah, sorry, I forgot about branches 🥲 Also it would be great if there was a warning in the readme with a link to the latest release's readme version, but no problem

PR #193, opened 2024-12-20
cache binaries in CI, set release title

AndrewChubatiuk:
- set release name to a tag name
- install binaries to $(REPO)/bin to be able to cache them in CI


dmitryk-dk (approved):
LGTM!

PR #203, opened 2025-01-17
issue-237: fix query type switching when creating alerts in Grafana

dmitryk-dk:
Made a port of this PR: https://github.com/VictoriaMetrics/victoriametrics-datasource/pull/246

Should fix the issue with the alerting panel in the Grafana


Loori-R (approved):
LGTM!

PR #213, opened 2025-01-24
backend: change INFO severity to DEBUG

hagen1778:
With this change we reduce amount of logs printed by Grafana. If needed, user can switch to DEBUG level to get everything logged by the plugin.

Based on plugin review recommendations.


dmitryk-dk (approved):
LGTM!

Issue #188, opened 2024-12-20
Victorialogs as a source for annotations 


opentokix:
I am trying to use Victorialogs Datasource as a source for annotations. Using python and requests to send a POST object to the jsonlines endpoint. I get annotations to work, but I don't understand what I need to do to get tags to work and how to construct my object/query-params to get labels (other) to work. And also have been unable to find this in any documentation. (I can write that doc in a PR if I just get some details how this is intended to work). 

I am using v0.13.2 of writing of this issue. 

![image](https://github.com/user-attachments/assets/4f97197c-a606-49e9-ba4a-a4f17a5bc4b1)

Here is the relevant python code for how I create the object. 

```python 

    stream_fields = "stream,level,invoker,system,type_of"
    uri = f"{scheme}://{host}:{port}/insert/jsonline?_stream_fields={stream_fields}"  # noqa E501
    headers = {
        "Content-Type": "application/stream+json",
        "Authorization": f"Bearer {token}"
    }
    doc = {"level": "info",
           "invoker": invoker,
           "system": system,
           "stream": "annotations",
           "type": type_of,
           "_msg": message,
           "tags": tags if tags else [],
           }
    try:
        response = requests.post(uri, headers=headers, data=json.dumps(doc))

```

This is the annotation I have gotten to work so far: 

![image](https://github.com/user-attachments/assets/915704e2-fa1b-4ebd-b5f3-74228f82af45)

It is the "Object object" I am wondering about.

dmitryk-dk:
@Loori-R could you take a look please?

opentokix:
Any progress on this?

dmitryk-dk:
> Any progress on this?

Hi! Will try to check it today

dmitryk-dk:
Hi @opentokix ! Could you share, please, the message, invoker, system, type_of variables?

opentokix:
I would like to have the mouse over more human readable. 

Lets say I send in a object whith the info of username, version and tags or something like that. 

 I want the mouse over to have that info only. And not a json-string. But it should be customizable

dmitryk-dk:
> I would like to have the mouse over more human readable.
> 
> Lets say I send in a object whith the info of username, version and tags or something like that.
> 
> I want the mouse over to have that info only. And not a json-string. But it should be customizable.
> 
[...]

PR #253, opened 2025-03-07
Transform labels to string for annotations in the dashboard

dmitryk-dk:
When the user tries to use labels in the annotations, they always transform to the [Object object] because it has Object representation.
Convert those labels to string representation.
Behavior for the Explore app stays the same for any request

Screenshot 2025-03-07 at 16 13 43

In the dashboard, we can construct annotations via the labels
Screenshot 2025-03-07 at 16 14 45

Screenshot 2025-03-07 at 16 15 03

Related issue: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/188

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,6 +17,8 @@ jobs:
       packages: write
       contents: write
       pages: write
+      attestations: write
+      id-token: write
     steps:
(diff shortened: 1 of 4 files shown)


Loori-R (approved):
LGTM!

Issue #249, opened 2025-02-28
log time shown in grafana Logs volume is incorrect


imcoder-cloud:
the logs time shown in grafana Logs volume is incorrect.
the query result of logs time range is 2025-02-28 04:54:40 ~ 2025-02-28 04:55:46, but it shown in grafana is around 2025-02-28 02:30 ~ 2025-02-28 07:00 like below

<img width="1196" alt="Image" src="https://github.com/user-attachments/assets/c2848b3d-a0c0-462d-81e2-e748da5117e2" />

Loki is shown like below

<img width="1138" alt="Image" src="https://github.com/user-attachments/assets/418869ef-70e4-4b93-b0cc-5dc4ab4ba591" />

And the log level does not reflected in Grafana Logs volume area, Loki are reflected correct.

victorialogs version: v1.14.0
datasource version

<img width="1203" alt="Image" src="https://github.com/user-attachments/assets/5b06b340-709a-44d6-a374-9cfc62aecb00" />

dmitryk-dk:
Hi @imcoder-cloud ! Could you please check the response from the api of the victorialogs and share it here? Is the response the same?

imcoder-cloud:
Hi @dmitryk-dk below is response while using Grafana
[response1.json](https://github.com/user-attachments/files/19037859/response1.json)
[response2.json](https://github.com/user-attachments/files/19037860/response2.json)

below response is using VMUI

[response3.json](https://github.com/user-attachments/files/19037917/response3.json)
[response4.json](https://github.com/user-attachments/files/19037916/response4.json)

dmitryk-dk:
> Hi [@dmitryk-dk](https://github.com/dmitryk-dk) below is response while using Grafana [response1.json](https://github.com/user-attachments/files/19037859/response1.json) [response2.json](https://github.com/user-attachments/files/19037860/response2.json)
> 
> below response is using VMUI
> 
> [response3.json](https://github.com/user-attachments/files/19037917/response3.json) [response4.json](https://github.com/user-attachments/files/19037916/response4.json)

Thank you! But can you get the response via the curl from the victorialogs API? 
I will check the responses you shared

imcoder-cloud:
Hi @dmitryk-dk the response from VMUI is called victorialogsAPI

<img width="698" alt="Image" src="https://github.com/user-attachments/assets/9fddcf98-9674-49ad-b4b5-947b45ea807d" />

[...]

PR #259, opened 2025-03-13
fix the bug with incorrect bar display on the Logs Volume chart

Loori-R:
Fixed bar rendering on the Logs Volume chart. The chart is now generated for each timestep, adding bars with zero values if no data is available. This prevents bar stretching.

Related issue: #249 

| Before | After |
|--------|-------|
| image | image |


dmitryk-dk (approved):
LGTM!

Issue #254, opened 2025-03-09
datasource cannot handle space in stream fields (VMUI does)


xreip:
Hi,

I've ran into a situation where a third party application was sending jsons with spaces in the keys (which is terrible, i agree). 
While using VMUI there's no problem to visualize those logs. Using Grafana, there's instantly an error.

I've been able to replicate the behavior with a dummy app :

![Image](https://github.com/user-attachments/assets/6406a9f6-3e41-40f3-8993-fe84331c7078)

![Image](https://github.com/user-attachments/assets/3e8da6b8-76d7-4dd2-940e-0516b1972fd9)

![Image](https://github.com/user-attachments/assets/8289efbb-9f41-4b07-98f3-f9d8bda2fe51)

![Image](https://github.com/user-attachments/assets/6b62b37e-8931-4640-b90c-1e2ad0dcc05e)

Setup :

- x86
- VL DS 0.15
- Grafana 11.1.3
-using otel with http headers : 
```
...
    headers:
      VL-Stream-Fields: Dino Species,transactions_type,buyer
```

There's multiple ways to normalize keys before ingestion in VL but since VMUI seems to handle it, i prefer to let you know.

Thanks

Loori-R:
@dmitryk-dk, could you take a look? It seems like the issue is on the backend side.

dmitryk-dk:
Hi @xreip ! Is it possible to get the response from the VictoriaLogs API in JSON format? I want to check what the stream label looks like.

dmitryk-dk:
Hi @xreip ! The issue was fixed in the [v0.16.1](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.16.1) release.

PR #266, opened 2025-03-19
fix parsing label names with '/' inside

dmitryk-dk:
Fixed issue with `/` in the label names. 

For example 

```{kubernetes.host="host1",kubernetes.labels.app.kubernetes.io/instance="123",kubernetes.labels.app.kubernetes.io/name="vmagent",kubernetes.namespace_name="monitoring"}```

All parsing functions are unexported, so I just copied part of the logic which is responsible for the stream parsing.

Related issue: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/265

### CODE REVIEW

File: pkg/utils/stream_fields_parser.go [hunk 1]
```diff
@@ -0,0 +1,395 @@
+package utils
```
hagen1778, line 1: I don't think you need to have this parser. This parser is for logsql parsing, support logical operations, regexes, etc. 
I think stream fields just can be parsed as simple as comma-separated list of key-value pairs enclosed in curly braces: `{channel_id=\"CGZF1H6L9\",channel_name=\"general\"}`.

dmitryk-dk: updated parser simplified a bit of logic. Most of the cases are covered by VictoriaLogs. Just parse and check a bit of the correct content of the `_stream` field

Loori-R (approved):

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/VictoriaMetrics/victorialogs-datasource
 go 1.24.1
 
 require (
+	github.com/VictoriaMetrics/VictoriaMetrics v1.113.0
```
hagen1778, line 6: Please run `go mod tidy && go mod vendor` to remove dependency that you don't use.

pkg/utils/stream_fields_parser.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// ParseStreamFields collect labels values pair from the _stream field
+func ParseStreamFields(streamFields string) ([]StreamField, error) {
+	if streamFields == "" {
+		return nil, nil
+	}
+
+	if !strings.HasPrefix(streamFields, "{") {
+		return nil, fmt.Errorf("_stream field must start with '{' or '}'")
```
hagen1778, line 22:
```suggestion
		return nil, fmt.Errorf("_stream field must start with '{'")
```

pkg/utils/stream_fields_parser.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	if !strings.HasSuffix(streamFields, "}") {
+		return nil, fmt.Errorf("_stream field must end with '}'")
+	}
+	if strings.HasSuffix(streamFields, `""}`) {
```
hagen1778, line 27: weird check, the quote could be anywhere before `}`

pkg/utils/stream_fields_parser_test.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+		{
+			name:         "incorrect value field without quotes",
+			streamFields: `{b=a}`,
+			want:         []StreamField(nil),
+			wantErr:      true,
+		},
+		{
+			name:         "empty label field in quotes",
+			streamFields: `{""="b"}`,
```
hagen1778, line 41: have you checked if vlogs can have `""` as stream key?

pkg/utils/stream_fields_parser_test.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+		{
+			name:         "incorrect label field",
+			streamFields: `{=a"}`,
+			want:         []StreamField(nil),
+			wantErr:      true,
+		},
+		{
+			name:         "incorrect label field with a quote after",
+			streamFields: `{"b""="a"}`,
```
hagen1778, line 53: have you checked if vlogs can have `b""` as stream key?

pkg/utils/stream_fields_parser_test.go
```diff
@@ -0,0 +1,106 @@
...
+					Label: "a",
+					Value: "b",
+				},
+				{
+					Label: "c",
+					Value: "e",
+				},
+			},
+			wantErr: false,
+		},
```
hagen1778, line 92: would you mind testing the following cases?
* { a = "b",  c="d"} // wantErr:false
* {a="b,c", d="e"} // wantErr:false
* {a="b=c",d="e"} // wantErr:false
* {a.b.c="d"} // wantErr:false
* {a.b,c,d="e"}// wantErr:false

hagen1778 (changes requested):

go.mod
dmitryk-dk: updated dependecies

pkg/utils/stream_fields_parser.go
dmitryk-dk: fixed

pkg/utils/stream_fields_parser.go
dmitryk-dk: removed this check and test.

pkg/utils/stream_fields_parser_test.go
dmitryk-dk: removed this test

pkg/utils/stream_fields_parser_test.go
dmitryk-dk: added tests and added additional test to cover some interesting label values combinations. All works well

pkg/utils/stream_fields_parser_test.go
dmitryk-dk: removed this tes

pkg/utils/stream_fields_parser.go
```diff
@@ -0,0 +1,64 @@
...
+	if len(streams) == 0 {
+		return nil, fmt.Errorf("_stream fields must contain at least one '{' or '}'")
+	}
+	labelValuesPairs := strings.Split(streams, "\",")
+
+	stf := make([]StreamField, 0, len(labelValuesPairs))
+	for _, labelValuePair := range labelValuesPairs {
+		labelValuePair = strings.TrimSpace(labelValuePair)
+		if labelValuePair[0] == '"' || labelValuePair[0] == '`' {
+			return nil, fmt.Errorf("label can not start with quote: %q", labelValuePair)
```
hagen1778, line 39:
```suggestion
			return nil, fmt.Errorf("stream name can not start with quote: %q", labelValuePair)
```

pkg/utils/stream_fields_parser.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+		fields := strings.SplitN(labelValuePair, "=", 2)
+		if len(fields) != 2 {
+			return nil, fmt.Errorf("incorrect label value pair %q in _stream field must have label=\"value\" format", labelValuePair)
```
hagen1778, line 43:
```suggestion
			return nil, fmt.Errorf("_stream field %q must have `label=\"value\"` format", labelValuePair)
```

pkg/utils/stream_fields_parser.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		label := strings.TrimSpace(fields[0])
+		if len(label) == 0 {
+			return nil, fmt.Errorf("label values pair %q must have label", labelValuePair)
```
hagen1778, line 48:
```suggestion
			return nil, fmt.Errorf("_stream field %q must have non-empty key", labelValuePair)
```

pkg/utils/stream_fields_parser.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		value := strings.TrimSpace(fields[1])
+		if len(value) == 0 || strings.EqualFold(value, `"`) {
+			return nil, fmt.Errorf("label values pair %q must have value", labelValuePair)
```
hagen1778, line 53:
```suggestion
			return nil, fmt.Errorf("_stream field %q must have non-empty value", labelValuePair)
```

hagen1778 (dismissed):
LGTM.
I would move vendor update to the different PR, as it has no relation to the current change.

dmitryk-dk:
> Related ticket #254 @dmitryk-dk please make sure it is covered as well.

The problem described in the issue is not related to this PR, it should be fixed in a separate PR.

hagen1778 (approved):

Issue #237, opened 2025-02-13
Ad Hoc filters autocomplete


konstantin-kornienko:
For now, Grafana says that Ad hoc filtering is not supported by the datasource, but in fact it works.

Is it possible to implement autocomplete for the filed values int the filter? It would be mega-helpful!

![Image](https://github.com/user-attachments/assets/9cd0d42f-8e1a-4d02-bd1e-ec7c74e59931)

dmitryk-dk:
@Loori-R could you take a look at this issue?

igorhrcek:
Any update on this issue?

Loori-R:
We are currently working on this issue.

hagen1778:
@konstantin-kornienko the fix was merged in https://github.com/VictoriaMetrics/victorialogs-datasource/pull/273 and will be available in the next release.

dmitryk-dk:
Hi @konstantin-kornienko , @igorhrcek ! The issue was fixed in the [v0.16.1](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.16.1) release. Close issue as solved. Feel free to reopen the issue if you find any problem

konstantin-kornienko:
@dmitryk-dk , @hagen1778 thanks, very appreciated! Tested, autocomplete is now working like a charm.

PR #273, opened 2025-03-27
fix: autocomplete for Ad Hoc filter 

Loori-R:
Fix autocomplete for Ad Hoc filter values.

Related issue: #237

image


hagen1778 (approved):
LGTM

Issue #255, opened 2025-03-09
Absent needed field values in Grafana variable


stefkkkk:
Hi, am I doing something wrong or it's a bug?

I want to extract some extra fields from query in Grafana variable, for this I'm using such query:
`cluster: "eks-prod-eu-west-3"  AND namespace: "ingress-nginx" | unpack_json | keep _time,upstream_name,ingress_host`

But them upstream_name and ingress_host fields are not appearing in Field Values

![Image](https://github.com/user-attachments/assets/d330a75d-1112-4dab-9fd9-1ffbec971507)

![Image](https://github.com/user-attachments/assets/e143fb4f-4c80-4343-8806-2540f9b07d87)

In compare, this fields are available if I switched to Field Names

![Image](https://github.com/user-attachments/assets/a7b7ce55-6135-422a-be6d-534da046e1f3)

How to get correctly needed fields values?

tiny-pangolin:
I believe this a duplicate of#226. The API used to get values doesn't support extracted field values right now. The way I worked around this was to parse logs using a collector. This approach also as the benefits of improving query performance since you do not have to parse the fields at query time

stefkkkk:
> I believe this a duplicate of[#226](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/226). The API used to get values doesn't support extracted field values right now. The way I worked around this was to parse logs using a collector. This approach also as the benefits of improving query performance since you do not have to parse the fields at query time

I think let it be the problems of users, who will use it in way like mine, I'm about performance. But that will be very useful, are any plans to do that, or not planning to implement completely?

tiny-pangolin:
I'm not sure if it is possible for the field values API to return fields generated by a query @valyala would probably be the best person to answer that question. 

A note about performance. I agree that users should be able to decide how to query their logs, but it is more performant to parse logs at ingestion time rather than query time since the query would have to proccess each message that matches the query each time a dashboard is loaded rather than 1 time during ingestion before looking for matching field values. Along with field values results being theoretical faster if it was possible, we have seen query results return twice as fast when looking for a specific field immediately rather than having to parse a message before looking for field values [as seen in this blog](https://rtfm.co.ua/en/vector-dev-introduction-aws-s3-logs-and-integration-with-victorialogs/)

stefkkkk:
[...]

PR #276, opened 2025-03-31
fix: correct variable field values loading in Grafana 

Loori-R:
1. Fixed loading of field values for Grafana variables.
2. Fixed query filter behavior during field values loading.
3. Added a loading indicator and error message when the request fails.

Related issue: #255

### DIFF
--- a/src/language_provider.ts
+++ b/src/language_provider.ts
@@ -12,14 +12,6 @@ interface FetchFieldsOptions {
   limit?: number;
 }
 
-interface FieldsRequestParams {
-  query: string;
-  start: number;
-  end: number;
-  limit?: number;
-  field?: string;
-}
-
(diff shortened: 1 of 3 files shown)


src/language_provider.ts
```diff
@@ -59,23 +51,29 @@ export default class LogsQlLanguageProvider extends LanguageProvider {
 
   async getFieldList(options: FetchFieldsOptions): Promise<FieldHits[]> {
     if (options.type === FilterFieldType.FieldValue && !options.field) {
+      console.warn('getFieldList: field is required for FieldValue type');
```
dmitryk-dk, line 54: Do we need this warn?

dmitryk-dk (approved):
LGTM! nit added small comment

PR #281, opened 2025-04-04
Enable freebsd amd64 build

AlexanderThaller:
The build works for freebsd amd64 so it should be enabled so that the plugin can be used under FreeBSD AMD64.

### DIFF
--- a/Magefile.go
+++ b/Magefile.go
@@ -22,7 +22,10 @@ func Build() {
 	linuxS390 := func () error {
 		return b.Custom("linux", "s390x")
 	}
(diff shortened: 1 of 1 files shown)


dmitryk-dk (commented):
Hi @AlexanderThaller ! Does Grafana support this build as well or you are using custom build?

AlexanderThaller:
I'm using the native ports for freebsd for grafana and victorialogs if thats what you mean.

https://www.freshports.org/www/grafana/
https://www.freshports.org/net-mgmt/victoria-logs/

I installed the victorialogs-datasource through the grafana webinterface which worked but there was no binary for freebsd-amd64 and adding it manually does not work of course because that will lead to signature mismatches.

dmitryk-dk (approved):
LGTM!

PR #301, opened 2025-05-20
add support for configuring log level using custom rules

Loori-R:
### Summary

This pull request adds a **"Log level rules"** section to the datasource configuration screen. It allows users to define custom rules for dynamically assigning log levels based on field values.

### How it works

Click **"Add rule"** to create a new rule. Each rule consists of the following elements:

1. **Enable switch** – toggle to enable or disable the rule.
2. **Field name** – the log field the rule applies to.
3. **Operator** – condition type:
   `Equals`, `Not equal`, `Matches regex`, `Less than`, `Greater than`.
4. **Value** – the value to compare against.
5. **Log level** – level to assign if the rule matches:
   `critical`, `warning`, `error`, `info`, `debug`, `trace`, `unknown`.
6. **Delete button** – removes the rule.

After adding or editing rules, make sure to click "Save & test" to apply the updated configuration.

### Rule priority

If multiple rules match, the **first one** in the list takes precedence.

image


dmitryk-dk (commented):
@Loori-R , could you please fix backend part, because of linter

Loori-R:
> @Loori-R , could you please fix backend part, because of linter

Fixed

src/datasource.ts
```diff
@@ -391,14 +397,16 @@ export class VictoriaLogsDatasource
...
         const step = Math.ceil(totalSeconds / LOGS_VOLUME_BARS) || "";
 
+        const fields = this.logLevelRules.filter(r => r.enabled).map(r => r.field)
+        const uniqFields = Array.from(new Set(fields));
+
         return {
           ...query,
           step: `${step}s`,
-          field: HITS_BY_FIELD,
+          field: uniqFields,
```
dmitryk-dk, line 409: I think we should rename JSON param to fields because it is an array now

pkg/plugin/query.go
```diff
@@ -53,7 +53,7 @@ type Query struct {
 	IntervalMs   int64     `json:"intervalMs"`
 	MaxLines     int       `json:"maxLines"`
 	Step         string    `json:"step"`
-	Field        string    `json:"field"`
```
dmitryk-dk, line 56: it is better to name the JSON field as `fields` because it is a slice

dmitryk-dk (commented):
added comments

dmitryk-dk (dismissed):
LGTM!

dmitryk-dk (approved):

arturminchukov (approved):

Issue #302, opened 2025-05-26
VictoriaLogs: Cannot set Grafana alert rule by logs count


motoki317:
### Describe the bug

When using VictoriaLogs datasource to setup an alert rule in Grafana, while I expect that I'll be able to setup log count alerts for each namespace in a single Alert Rule, I get the following error:

> [sse.readDataError] [A] got error: input data must be a wide series but got type not (input refid)

<img width="1183" alt="Image" src="https://github.com/user-attachments/assets/6d38ea94-4ff9-4d6e-99db-ceee24b424a8" />

However, when I don't have the grouping `by (namespace)` in stats pipe, it works without a problem.

<img width="1192" alt="Image" src="https://github.com/user-attachments/assets/4914398e-0724-423a-ab48-afb5479735d2" />

I'm not sure if this has to do with VictoriaLogs datasource plugin, or Grafana itself.

### To Reproduce

1. Setup VictoriaLogs (cluster), and Grafana with VictoriaLogs datasource plugin.
2. Create an alert rule on Grafana with something like the following: `` _time:5m ~`WARN|ERROR` | stats by (namespace) count() logs ``
3. Observe the error.

### Version

`victoria-logs-20250425-174411-tags-v1.21.0-victorialogs-0-g9d82de9385`

- VictoriaLogs cluster helm chart v0.0.2
- VictoriaLogs version v1.21.0
- Grafana v12.0.0+security-01 (6475e5a4a4)
- VictoriaLogs datasource plugin 0.16.3

### Logs

There doesn't seem to be any error logs neither in Grafana nor VictoriaLogs components (vlinsert, vlselect, vlstorage).

motoki317:
@AndrewChubatiuk Hi, thank you for a quick response.
Yes, I have tried the same query both in Grafana Explore and in VictoriaLogs UI.
In Grafana Explore, I'm getting my expected results in all formats (raw logs, range, and instant).

<img width="1422" alt="Image" src="https://github.com/user-attachments/assets/680d0cb0-ef0e-41ad-befe-70ba3d00ad5f" />
<img width="1438" alt="Image" src="https://github.com/user-attachments/assets/731197fa-575c-4e80-95d4-cbf5cb3f5e20" />
<img width="1436" alt="Image" src="https://github.com/user-attachments/assets/3e009874-4e5a-4d11-af80-da3f9f8841fa" />

There's also no errors in VictoriaLogs ui.

[...]

PR #303, opened 2025-05-27
fix error when response detected as not a wide series

dmitryk-dk:
Grafana tries to detect the type of the frame, and we need to specify that type for the instant queries in the alerting tab

Related issue: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/302

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## tip
 
+* BUGFIX: fix an issue when Grafana decides that the response is not a wide series and shows the error "input data must be a wide series but got type not (input refid)" . See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/302).
+
 ## v0.16.3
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## tip
 
+* BUGFIX: fix error when response detected as not a wide series for the instant queries. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/302).
```
hagen1778, line 5: @dmitryk-dk could you please re-phraze it in a way that is more clear to end user?
For example, `fix error "input data must be a wide series" on Grafana alerting page`. This would make it clear to the user that fix is for alerts. And "input data must be a wide series" mentioning would make it searchable if user googles the error.
Please also try using format [#302](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/302) instead of [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/302), as it removes constantly repeating `this issue` phrases.

Loori-R (dismissed):

Issue #244, opened 2025-02-25
Filter hints do not narrow the scope of possible values


tropnikovvl:
In loki datasource automatically narrows down possible values ​​based on previously selected labels, in vmlogs - no.

How it looks in screenshots

Loki - shows values ​​only from the `test` cluster

![Image](https://github.com/user-attachments/assets/9ebe1b00-c71e-437a-9d56-f93478c1476c)

VMlogs - shows all values

![Image](https://github.com/user-attachments/assets/830fbc85-2e6d-4d1c-a31a-25adc1f34412)

zkl94:
any news on this issue? it feels like this should be baseline functionality instead of afterthought

Loori-R:
> any news on this issue? it feels like this should be baseline functionality instead of afterthought

We'll take a look at this issue in the coming days.

stefan-korchahin:
> Hi [@zkl94](https://github.com/zkl94) ! The feature was released and is available in the release [v0.18.0](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.18.0). If you find any issue, please reopen this one.

seems it still not working

![Image](https://github.com/user-attachments/assets/15eb21ff-3080-46ed-91a7-0f75d7a4a805)

stefan-korchahin:
> > > Hi [@zkl94](https://github.com/zkl94) ! The feature was released and is available in the release [v0.18.0](https://github.com/VictoriaMetrics/victorialogs-datasource/releases/tag/v0.18.0). If you find any issue, please reopen this one.
> > 
> > 
> > seems it still not working
> > ![Image](https://github.com/user-attachments/assets/15eb21ff-3080-46ed-91a7-0f75d7a4a805)
> 
> Hi [@stefan-korchahin](https://github.com/stefan-korchahin) ! Can you do a hard reload with cache clear (Ctrl+Shift+R or Cmd+Shift+R) and check if the issue is present?

oh that helps, everything works, thank you

PR #309, opened 2025-06-10
Improve query builder: narrow field/value options based on previous filters

Loori-R:
When using the `AND` operator in the query builder, the list of available fields and values is now dynamically narrowed based on the previously selected filters.

Related issue: #244

image
image


dmitryk-dk (approved):
LGTM!

arturminchukov (approved):

PR #316, opened 2025-06-18
fix an issue where line charts were incorrectly connecting data points across missing (null) values despite the "Connect null values" panel setting being set to "Never"

arturminchukov:
1. Added an `interval` config param to time field to fix an issue where line charts were incorrectly connecting data points across missing (null) values despite the "Connect null values" panel setting being set to "Never"

Connect null values: never
image

Connect null values: always
image

2. Updated readme with instructions how to run backend plugin in debug mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 * FEATURE: improve query builder – when using AND operator, field and value selection is now narrowed. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/244).
 * FEATURE: add a section to the datasource settings for tenant configuration. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/307).
 
+* BUGFIX: fix an issue where line charts were incorrectly connecting data points across missing (null) values despite the "Connect null values" panel setting being set to "Never". See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/311).
+
 ## v0.17.0
 
 * FEATURE: add support for configuring log level using custom rules. See [this issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/294).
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -300,6 +301,36 @@ yarn dev
 # build the plugin for production to the `victoriametrics-logs-datasource` folder and zip build
 yarn build:zip
 ```
+#### 2. How to run a debugger for a backend plugin:
```
dmitryk-dk, line 304: I would suggest using this debug behavior 
https://st-g.de/2023/10/grafana-plugin-debugging-on-apple-silicon

go.mod
```diff
@@ -101,3 +101,5 @@ require (
 	gopkg.in/fsnotify/fsnotify.v1 v1.4.7 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+tool github.com/magefile/mage
```
dmitryk-dk, line 105: why do we need this in the go.mod?

dmitryk-dk (commented):
Please check comments and the tests on backend part

go.mod
dmitryk-dk: If someone wants to use it, they can install it. We do not need this in the project as a dependency

arturminchukov: anyway `mage` will be installed by [makefile](https://github.com/VictoriaMetrics/victorialogs-datasource/blob/ed3eb4681abc4e8dcc31a6543485d8a8b2965b8a/Makefile#L112-L115)
So, I would like to suggest move it to `go.mod` tool. And change to `go tool mage` into makefile. But anyway will remove these changes from this PR. And if you agree, will do it in another PR.

arturminchukov:
> [Backend tests are failing](https://github.com/VictoriaMetrics/victorialogs-datasource/actions/runs/15727310302/job/44320029018?pr=316) — could someone please take a look?

Thanks, fixed it

go.mod
dmitryk-dk: We do not need this tool to build the go files. We used it to sign the plugin because of the Grafana rules.

dmitryk-dk (approved):
LGTM!

Issue #298, opened 2025-05-08
ad-hoc filtering is applied incorrectly with extract pipe


pgassmann:
When there is an `| extract` pipe, the ad hoc filters are not correctly applied yet. the filter is added in front of the `| stats` pipe, but after the `| extract` pipe, which breaks the query.

Related to 
* ad-hoc filtering is applied incorrectly #135
* fix incorrect application of ad-hoc filters #144 

query:
```
host:in($host) and compose_project:in($compose_project) and compose_service:in($compose_service) 
  and $log_query
  | stats by (host) count()
```

$log_query is `REST CALL by" | extract "REST CALL by <restuser>:`

error:
```
error: cannot parse query [host:in(prod01) and compose_project:in(example-prod) and compose_service:in(portal) and "REST CALL by" | extract "REST CALL by <restuser>:" AND container_name:!="test" | stats by (host) count()]: unexpected token after [extract "REST CALL by <restuser>:"]: "AND"; expecting '|' or ')'; context: [CALL by" | extract "REST CALL by <restuser>:" AND]
```

 ![Image](https://github.com/user-attachments/assets/ab76429f-3f47-4a21-a459-1500b953c390)

pgassmann:
Instead of modifying the query string, the ad hoc filters could be passed as extra_filters.

From: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/293#issuecomment-2831359370 by @valyala 
> VictoriaLogs accepts multiple `extra_filters` query args across all the supported querying APIs starting from [v1.21.0-victorialogs](https://docs.victoriametrics.com/victorialogs/changelog/) according to [these docs](https://docs.victoriametrics.com/victorialogs/querying/#extra-filters).

PR #322, opened 2025-06-24
fix an issue where ad-hoc filtering is applied incorrectly with extract pipe

arturminchukov:
Related issue: #298 

Fixed an issue where ad-hoc filtering is applied incorrectly with extract pipe via moving ad-hoc filters from query to extra_filters query arg


Loori-R (dismissed):
LGTM! Thanks!

PR #324, opened 2025-06-24
update readme with the debug instructions for the backend plugin

arturminchukov:
update readme with the debug instructions for the backend plugin


dmitryk-dk (commented):
Overall looks good!
Could you please provide the description for the PR?

PR #329, opened 2025-07-04
Implement Grafana Query Builder

archef2000:
Add support for all operations of LogsQL in the Grafana Query Builder.
@Loori-R

#48

### DIFF
--- a/src/components/QueryEditor/QueryBuilder/Editors/ExactValueEditor.tsx
+++ b/src/components/QueryEditor/QueryBuilder/Editors/ExactValueEditor.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+
+import { QueryBuilderOperationParamEditorProps } from '@grafana/plugin-ui';
+import { Combobox } from '@grafana/ui';
+
+import { getFieldValueOptions } from './utils/editorHelper';
+
+export default function ExactValueEditor(props: QueryBuilderOperationParamEditorProps) {
+  const { onChange, index, value, operation } = props;
+
+  return (
+    <Combobox<string>
+      options={()=> getFieldValueOptions(props, operation.params[0] as string)}
+      value={value as string}
+      onChange={(v) => onChange(index, v.value)}
+      width="auto"
+      maxWidth={30}
+      minWidth={10}
+      createCustomValue
+    />
+  );
+}
--- a/src/components/QueryEditor/QueryBuilder/Editors/FieldAsFieldEditor.tsx
+++ b/src/components/QueryEditor/QueryBuilder/Editors/FieldAsFieldEditor.tsx
@@ -1,0 +1,80 @@
+import React, { useState } from 'react';
+
+import { QueryBuilderOperationParamEditorProps } from '@grafana/plugin-ui';
+import { InlineField, Stack, Combobox } from '@grafana/ui';
+
+import { isValue, quoteString, getValue } from '../utils/stringHandler';
+import { splitString } from '../utils/stringSplitter';
+
+import { getFieldNameOptions } from './utils/editorHelper';
+
+export default function FieldAsFieldEditor(props: QueryBuilderOperationParamEditorProps) {
+  const { value, onChange, index } = props;
+
+  const str = splitString(value as string);
+  let parsedFromField = "";
+  let parsedToField = "";
+  if (str.length === 3) {
+    if (isValue(str[0])) {
+      parsedFromField = getValue(str[0]);
+    }
+    if (str[1].type === "space" && str[1].value === "as") {
+      if (isValue(str[2])) {
+        parsedToField = getValue(str[2]);
+      }
+    }
+  }
+
+  const [fromField, setFromField] = useState<string>(parsedFromField);
+  const [toField, setToField] = useState<string>(parsedToField);
+
+  const updateValue = (fromField: string, toField: string) => {
+    let value = "";
+    if (fromField.trim() === "") {
+      value += "\"\"";
+    } else {
+      value += quoteString(fromField.trim());
+    }
+    value += " as ";
+    if (toField.trim() === "") {
+      value += "\"\"";
+    } else {
+      value += quoteString(toField.trim());
+    }
+    onChange(index, value);
+  };
+
+  return (
+    <Stack>
+      <InlineField>
+        <Combobox<string>
+          createCustomValue
+          options={()=>getFieldNameOptions(props)}
+          onChange={(value) => {
+            setFromField(value.value);
+            updateValue(value.value, toField);
+          }}
+          value={fromField}
+          width="auto"
+          maxWidth={30}
+          minWidth={10}
+        />
+      </InlineField>
+      <div style={{ padding: '6px 0 8px 0px' }}>as</div>
+      <InlineField>
+        <Combobox<string>
+          createCustomValue
+          options={()=>getFieldNameOptions(props)}
+          onChange={(value) => {
+            setToField(value.value);
+            updateValue(fromField, value.value);
+          }}
+          value={toField}
+          width="auto"
+          maxWidth={30}
+          minWidth={10}
+        />
+      </InlineField>
+    </Stack>
+  );
+}
--- a/src/components/QueryEditor/QueryBuilder/Editors/FieldsEditorWithPrefix.tsx
+++ b/src/components/QueryEditor/QueryBuilder/Editors/FieldsEditorWithPrefix.tsx
@@ -1,0 +1,121 @@
+import React, { useState } from 'react';
+
+import { SelectableValue } from '@grafana/data';
+import { QueryBuilderOperationParamEditorProps } from '@grafana/plugin-ui';
+import { MultiSelect } from '@grafana/ui';
+
+import { quoteString, unquoteString } from '../utils/stringHandler';
+import { splitByUnescapedChar, SplitString, splitString } from '../utils/stringSplitter';
+
+import { getFieldNameOptions } from './utils/editorHelper';
+
+interface FieldWithPrefix {
+  name: string;
+  isPrefix: boolean;
+}
+
+export default function FieldsEditorWithPrefix(props: QueryBuilderOperationParamEditorProps) {
+  const { value, onChange, index } = props;
+
+  const str = splitString(value as string);
+  const parsedValues = parseInputValues(str);
+  const [values, setValues] = useState<FieldWithPrefix[]>(parsedValues);
+
+  const setFields = (values: FieldWithPrefix[]) => {
+    setValues(values);
+    const newValue = values.map((field) => {
+      if (field.isPrefix !== undefined) {
+        return field.isPrefix ? `${quoteString(field.name)}*` : `${quoteString(field.name)}`;
+      } else {
+        return quoteString(field as unknown as string);
+      }
+    }).join(', ');
+    onChange(index, newValue);
+  }
+
+  const togglePrefix = (index: number) => {
+    const field = values[index];
+    field.isPrefix = !field.isPrefix;
+    setFields(values);
+  };
+
+  const [state, setState] = useState<{
+    options?: SelectableValue<FieldWithPrefix>[];
+    isLoading?: boolean;
+  }>({});
+
+  return (
+    <MultiSelect<FieldWithPrefix>
+      openMenuOnFocus
+      onOpenMenu={async () => {
+        setState({ isLoading: true });
+        let options = await getFieldNameOptions(props);
+        const selectedNames = values.map(v => v.name);
+        options = options.filter((opt: SelectableValue<string>) => opt.value && !selectedNames.includes(opt.value));
+        setState({ options, isLoading: undefined });
+      }}
+      isLoading={state.isLoading}
+      allowCustomValue
+      noOptionsMessage="No labels found"
+      loadingMessage="Loading labels"
+      options={state.options}
+      value={values}
+      onChange={(values) => setFields(values.map((v) => v.value || v as FieldWithPrefix))}
+      formatOptionLabel={(option, { context }) => {
+        if (context === 'value') {
+          const field = option as FieldWithPrefix;
+          const handleToggle = (e: React.SyntheticEvent) => {
+            e.stopPropagation();
+            const idx = values.findIndex((v) => (v).name === field.name);
+            if (idx !== -1) {
+              togglePrefix(idx);
+            }
+          };
+          return (
+            <span
+              tabIndex={0}
+              style={{ cursor: 'pointer' }}
+              onMouseDown={handleToggle}
+            >
+              {formatFieldLabel(field)}
+            </span>
+          );
+        }
+        return <>{option.label}</>;
+      }}
+    />
+  );
+}
+
+const formatFieldLabel = (field: FieldWithPrefix): string => {
+  return field.isPrefix ? `${field.name} *` : field.name;
+};
+
+const parseValue = (value: SplitString[]): FieldWithPrefix => {
+  if (value.length === 0 || value[0].type === "bracket") {
+    return { name: '', isPrefix: false };
+  }
+  if (value[0].type === "quote") {
+    let isPrefix = false;
+    if (value.length > 1) {
+      isPrefix = (value[1].type === "space" && value[1].value === "*");
+    }
+    return { name: unquoteString(value[0].value), isPrefix };
+  } else {
+    let fieldValue = value[0].value;
+    if (fieldValue.endsWith('*')) {
+      return { name: fieldValue.slice(0, -1), isPrefix: true };
+    }
+    return { name: fieldValue, isPrefix: false };
+  }
+}
+
+const parseInputValues = (str: SplitString[]): FieldWithPrefix[] => {
+  let fields: FieldWithPrefix[] = [];
+  for (const field of splitByUnescapedChar(str, ',')) {
+    if (field.length > 0 && field[0].type !== "bracket") {
+      fields.push(parseValue(field));
+    }
+  }
+  return fields;
+};
--- a/src/components/QueryEditor/QueryBuilder/Editors/NumberEditor.tsx
+++ b/src/components/QueryEditor/QueryBuilder/Editors/NumberEditor.tsx
@@ -1,0 +1,32 @@
+import React from "react";
+
+import { QueryBuilderOperationParamEditorProps } from "@grafana/plugin-ui";
+
+export default function NumberEditor(props: QueryBuilderOperationParamEditorProps) {
+    const { value, onChange, index } = props;
+    const validateInput = (input: string): string => {
+        // allow numbers, -, +, ., KB, MB, GB, TB, Ki, Mi, Gi, Ti, K,M, G, T, KiB, MiB, GiB, TiB, _
+        const allowedPattern = /[^0-9\-+._KMGTkmgtiIBb]/g;
+        return input.replace(allowedPattern, '');
+    }
+    const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        const inputValue = e.target.value;
+        const sanitizedValue = validateInput(inputValue);
+        if (sanitizedValue !== inputValue) {
+            e.target.value = sanitizedValue;
+        }
+    };
+    const updateValue = (e: React.FocusEvent<HTMLInputElement>) => {
+        onChange(index, e.target.value.trim());
+    }
+    
+    return (
+        <input 
+            type="text"
+            defaultValue={value as string}
+            onInput={handleInputChange}
+            onBlur={updateValue}
+            style={{ width: '15ch' }}
+        />
+    );
+}
--- a/src/components/QueryEditor/QueryBuilder/Editors/SortedFieldsEditor.tsx
+++ b/src/components/QueryEditor/QueryBuilder/Editors/SortedFieldsEditor.tsx
@@ -1,0 +1,117 @@
+import React, { useState } from 'react';
+
+import { SelectableValue } from '@grafana/data';
+import { QueryBuilderOperationParamEditorProps } from '@grafana/plugin-ui';
+import { Icon, MultiSelect } from '@grafana/ui';
+
+import { quoteString, unquoteString } from '../utils/stringHandler';
+import { splitByUnescapedChar, SplitString, splitString } from '../utils/stringSplitter';
+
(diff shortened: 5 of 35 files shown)


Loori-R (changes requested):
Tested on Grafana `11.3.1`.

An error occurs when adding filters:
1) Open Explorer
2) Go to `Operations` → `Filters` → `Words`
This issue happens not only with **Words**, but also with other filters.

image

archef2000:
I am currently on 12.0.2 and everything is working will test version 11

archef2000:
The problem exists up to including v 11.4.6 and is fixed in 11.5.0.
The source problem is that the custom Field Editors get undefined as props

dk:
> It looks like the issue is with the `Combobox` component - it doesn't work correctly with older versions of Grafana.
> The `Select` component (which was its alternative) is now considered deprecated.
> 
> I think we should consider raising the minimum supported Grafana version.
> What do you think, @dk, @roman?
> 

In my experience query builders are fairly hard to use manually, - click click click hundred times, - I prefer plain old text for that. But what do I know, I am probably a wrong dk :)

Loori-R:
> In my experience query builders are fairly hard to use manually, - click click click hundred times, - I prefer plain old text for that. But what do I know, I am probably a wrong dk :)

Oops, looks like I tagged the wrong person 😅
Thanks a lot for sharing your thoughts - and sorry for the ping!

Loori-R:
* There is no "old" `Combobox` - it was introduced as a new component and is currently marked as **Alpha**. [View source](https://github.com/grafana/grafana/blob/main/packages/grafana-ui/src/components/Combobox/Combobox.tsx#L112-L116)

* The `Select` component is marked as **Deprecated**. [View source](https://github.com/grafana/grafana/blob/main/packages/grafana-ui/src/components/Select/Select.tsx#L13)

As an option, we could use `Select` for now since it’s more stable, and plan to migrate to `Combobox` in the future.
[Guide: Migrating from Select to Combobox](https://developers.grafana.com/ui/latest/index.html?path=/docs/forms-combobox--docs#migrating-from-select)

archef2000:
@Loori-R Can comfirm that with the Select component everything works.

Loori-R:
> @Loori-R Can comfirm that with the Select component everything works.

@archef2000 Would you mind replacing `Combobox` with `Select`?

archef2000:
Already on it

archef2000:
This query should give me the value types of `ClientPort` right?
`/select/logsql/field_values?query=_msg%3A*+%7C+uniq+by+ClientPort+%7C+block_stats+%7C+uniq+type&start=1751839200000&end=1752011999999&field=type`

query: `_msg:* | uniq by ClientPort | block_stats | uniq type`
field: `type`

dmitryk-dk:
> _msg:* | uniq by ClientPort | block_stats | uniq type

Sorry guys missed you question

query `_msg:* | uniq by ClientPort | block_stats` will return 

```
field: ClientPorttype: constvalues_bytes: 0bloom_bytes: 0dict_items: 0dict_bytes: 0rows: 1part_path: inmemory
```
the next filter will get only type so the query like 
`_msg:* | uniq by ClientPort | block_stats | uniq type` will return only `type: const`

archef2000:
I was in the assumption that /field_values will return the values of the field that are available after the query. Just noticed it in the value_type operation, but then I will have to change the logic for all. Should I change the getFieldList function with an additional type?

Loori-R:
Maybe we could just hardcode the supported value types (like `const`, `dict`, `string`, `int64`, `float64`, etc.) and allow users to enter a custom type if needed - that might be a simpler and more flexible solution.

src/modifyQuery.ts
```diff
@@ -34,50 +33,150 @@ const getFilterInsertValueForStream = (key: string, value: string, operator: str
...
+  visQuery.operations = newOps;
 
-  return removed;
-}
+  visQuery.operations = visQuery.operations.filter(op => {
+    if (op.id === VictoriaLogsOperationId.Logical) {
+      op.params[1] = removeLabelFromQuery(op.params[1] as string, key, value, operator, op.params[0] as string);
+    }
+    return true;
+  })
```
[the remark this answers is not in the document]
Loori-R: @archef2000, could you please take a look? This seems like a valid comment.

src/components/QueryEditor/QueryBuilder/QueryModeller.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { parseOperation, parseStatsOperation } from './utils/operationParser';
+import { splitByOperator, splitByUnescapedPipe, splitString } from './utils/stringSplitter';
+
+export function createQueryModellerWithDefaultField(defaultField: string, categegories: VictoriaLogsQueryOperationCategory[]) {
+  const queryModeller = new QueryModeller(new OperationDefinitions(defaultField).all(), categegories);
+  return queryModeller;
+}
+
+export function createQueryModellerForCategories(categegories: VictoriaLogsQueryOperationCategory[] = Object.values(VictoriaLogsQueryOperationCategory)) {
+  const queryModeller = new QueryModeller(new OperationDefinitions().all(), categegories);
```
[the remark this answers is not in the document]
Loori-R: @archef2000, can you check?

hagen1778:
> I think we should consider raising the minimum supported Grafana version.

I don't like it. According to plugin stats, 50% of users are using <v12  Grafana version:

image

dmitryk-dk:
> > I think we should consider raising the minimum supported Grafana version.
> 
> I don't like it. According to plugin stats, 50% of users are using <v12 Grafana version:
> 
> image

We decided to use the old Select component and stay with the version we have now

src/components/QueryEditor/QueryBuilder/utils/stringSplitter.test.ts
```diff
@@ -0,0 +1,256 @@
...
+  it("should handle complex strings with nested brackets and angled brackets", () => {
+    const input = `'hello' (world [nested "value")] key:value "key:value"`;
+    const expected = [
+      { type: "quote", value: "'hello'" },
+      {
+        prefix: "",
+        type: "bracket",
+        value: [
+          { type: "space", value: "world" },
+          { type: "bracket", prefix: "", value: [{ type: "space", value: 'nested' }, { type: "quote", value: '"value"' }], raw_value: "[nested \"value\")" },
```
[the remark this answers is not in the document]
archef2000: Yes the mismatch is intended to test for the range filter, etc.

src/components/QueryEditor/QueryBuilder/utils/operationParser.ts
```diff
@@ -0,0 +1,242 @@
...
+                fieldName = str[0].value;
+                str.shift();
+                str.shift();
+            }
+        } else if (str[0].type === "quote" && str[1].type === "colon" && str[1].value === "") { // "field1" :
+            fieldName = unquoteString(str[0].value);
+            str.shift();
+            str.shift();
+        }
+    }
```
Loori-R, line 76: Suggest extracting repeated checks into well-named variables and flattening the nested blocks to improve readability.
For example, you can also extract `str[0]` and `str[1]` into variables:


```javascript
const { type: typeFirst, value: valueFirst } = str[0] || {};
const { type: typeSecond, value: valueSecond } = str[1] || {};

const isColon = typeFirst === "colon";
const isSpaceColon = typeFirst === "space" && typeSecond === "colon" && valueSecond === "";
const isQuoteColon = typeFirst === "quote" && typeSecond === "colon" && valueSecond === "";

if (isColon) {
  fieldName = valueFirst;
  str.shift();
} else if (isSpaceColon && !isFilterFunction(valueFirst) && !isStatFunction(valueFirst)) {
  fieldName = valueFirst;
  str.shift();
  str.shift();
} else if (isQuoteColon) {
  fieldName = unquoteString(valueFirst);
  str.shift();
  str.shift();
}

```

src/components/QueryEditor/QueryBuilder/Editors/FieldAsFieldEditor.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <Select<string>
+          allowCustomValue={true}
+          allowCreateWhileLoading={true}
+          isLoading={state.loading}
+          onOpenMenu={handleOpenMenu}
+          options={state.options}
+          onChange={(value) => {
+            setFromField(value.value as string);
+            updateValue(value.value as string, toField);
+          }}
```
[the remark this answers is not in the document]
archef2000: solved for all occurrences I could find

src/components/QueryEditor/QueryBuilder/Editors/NumberEditor.tsx
```diff
@@ -0,0 +1,32 @@
...
+    }
+    
+    return (
+        <input 
+            type="text"
+            defaultValue={value as string}
+            onInput={handleInputChange}
+            onBlur={updateValue}
+            style={{ width: '15ch' }}
+        />
```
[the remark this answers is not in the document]
archef2000: Yes as the Grafana Input component does not respect the width style

src/components/QueryEditor/QueryBuilder/Editors/FieldAsFieldEditor.tsx
```diff
@@ -0,0 +1,96 @@
...
+    }
+    if (str[1].type === "space" && str[1].value === "as") {
+      if (isValue(str[2])) {
+        parsedToField = getValue(str[2]);
+      }
+    }
+  }
+
+  const [fromField, setFromField] = useState<string>(parsedFromField);
+  const [toField, setToField] = useState<string>(parsedToField);
```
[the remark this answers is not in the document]
archef2000: So this would be totally save.
```
  if (str.length === 3) {
    parsedFromField = getValue(str[0]);
    parsedToField = getValue(str[2]);
  }
```

src/components/QueryEditor/QueryBuilder/Operations.tsx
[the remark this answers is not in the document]
archef2000: Should I just create one helper file or split it up even more?

src/components/QueryEditor/QueryBuilder/Editors/ExactValueEditor.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <InlineField>
+      <Select<string>
+        allowCustomValue={true}
+        allowCreateWhileLoading={true}
+        isLoading={isLoading}
+        onOpenMenu={async () => {
+          setIsLoading(true);
+          setOptions(await getFieldValueOptions(props, operation.params[0] as string));
+          setIsLoading(false);
```
dmitryk-dk, line 23: i think we can define function and use it. Also please take a look on `onChange`

src/components/QueryEditor/QueryBuilder/Editors/FieldAsFieldEditor.tsx
```diff
@@ -0,0 +1,85 @@
...
+  return (
+    <Stack>
+      <InlineField>
+        <Select<string>
+          allowCustomValue={true}
+          allowCreateWhileLoading={true}
+          isLoading={isLoading}
+          onOpenMenu={handleOpenMenu}
+          options={options}
+          onChange={({ value = "" }) => {
```
dmitryk-dk, line 59: it is better to move this to separate handler

archef2000: I think it is clearer that way

src/components/QueryEditor/QueryBuilder/Editors/FieldsEditorWithPrefix.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const str = splitString(String(value || ""));
+  const parsedValues = parseInputValues(str);
+  const [values, setValues] = useState<FieldWithPrefix[]>(parsedValues);
+
+  const setFields = (values: FieldWithPrefix[]) => {
+    setValues(values);
+    const newValue = values.map((field) => {
+      if (field.isPrefix !== undefined) {
+        return field.isPrefix ? `${quoteString(field.name)}*` : `${quoteString(field.name)}`;
+      } else {
```
dmitryk-dk, line 29: nit. else is not needed

src/components/QueryEditor/QueryBuilder/Editors/FieldsEditorWithPrefix.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+const parseValue = (value: SplitString[]): FieldWithPrefix => {
+  if (value.length === 0 || value[0].type === "bracket") {
+    return { name: '', isPrefix: false };
+  }
+  if (value[0].type === "quote") {
+    const { type, value: secondValue } = value[1] || {};
+    const isPrefix = type === "space" && secondValue === "*";
+    return { name: unquoteString(value[0].value), isPrefix };
+  } else {
```
dmitryk-dk, line 104: nit. else not needed here

dmitryk-dk (commented):
I suggest a little cleanup of the PR with callbacks, because every component has a handler, but its usage is on the element.

src/components/QueryEditor/QueryBuilder/Editors/FieldsEditorWithPrefix.tsx
archef2000: It is needed for adding a field initially

dmitryk-dk:
```javascript
if (field.isPrefix !== undefined) {
        return field.isPrefix ? `${quoteString(field.name)}*` : `${quoteString(field.name)}`;
} 
return quoteString(field as unknown as string);
```
Will behave in the same way, but without additional branch

archef2000:
@dmitryk-dk
> I suggest a little cleanup of the PR with callbacks, because every component has a handler, but its usage is on the element.

May you be able to give examples.

dmitryk-dk:
> @dmitryk-dk
> 
> > I suggest a little cleanup of the PR with callbacks, because every component has a handler, but its usage is on the element.
> 
> May you be able to give examples.

```javascript
import React, { useState } from 'react';

import { SelectableValue } from '@grafana/data';
import { QueryBuilderOperationParamEditorProps, toOption } from '@grafana/plugin-ui';
import { InlineField, Select } from '@grafana/ui';

import { getFieldNameOptions } from './utils/editorHelper';

export default function FieldEditor(props: QueryBuilderOperationParamEditorProps) {
  const { value, onChange, index } = props;
  const [isLoading, setIsLoading] = useState(false);
  const [options, setOptions] = useState<SelectableValue<string>[]>([]);
  
  const onOpenMenu = async () => {
          setIsLoading(true);
          setOptions(await getFieldNameOptions(props));
          setIsLoading(false);
  }

 const handleOnChange = ({ value = "" }) => {
          onChange(index, value);
   }

  return (
    <InlineField>
      <Select<string>
        allowCustomValue={true}
        allowCreateWhileLoading={true}
        isLoading={isLoading}
        onOpenMenu={onOpenMenu}
        options={options}
        onChange={handleOnChange}
        value={toOption(String(value || ""))}
        width="auto"
      />
    </InlineField>
  );
}
```

Something like this

archef2000:
Could the onChange not just be in one line? Like I did in [ExactValueEditor.tsx](https://github.com/VictoriaMetrics/victorialogs-datasource/pull/329/commits/238b3b25928fc098cc2fce00b7be68f0d2439847#diff-5fa1811c841755f3f3a69db63886fe62719a7510163b9668a2225edf8f5b1e80)

And just make handlers for onOpenMenu callbacks?

archef2000:
Why are only filters respected in the field_values/field_names endpoints. When it is stated otherwise in the docs? @Loori-R @dmitryk-dk

Loori-R:
> Why are only filters respected in the field_values/field_names endpoints. When it is stated otherwise in the docs? @Loori-R @dmitryk-dk

The `/field_values` endpoint returns a list of unique values for a given field from the raw logs.
Just to clarify - could you provide an example of what you’re expecting to achieve with pipes in this context?

archef2000:
For the every filter I construct all operation before that one and get all (at this point) available field values, names, types.

Loori-R (dismissed):
Thanks for the pull request! LGTM!

archef2000:
I would like the operation field names/values recommendation to be correct. So an answer to my question would be appreciated to know if I need a different approach or if the implementation will be adjusted according to the docs.

Loori-R:
> Why are only filters respected in the field_values/field_names endpoints. When it is stated otherwise in the docs? @Loori-R @dmitryk-dk

@valyala could you help clarify this?

src/components/QueryEditor/QueryBuilder/Editors/SortedFieldsEditor.tsx
```diff
@@ -0,0 +1,114 @@
...
+              {formatFieldLabel(field)}
+              <Icon
+                name={field.isDesc ? 'arrow-down' : 'arrow-up'}
+                style={{ marginLeft: '4px', verticalAlign: 'middle' }}
+              />
+            </span>
+          );
+        }
+        return <>{option.label}</>;
+      }}
```
dmitryk-dk, line 88: this is huge callback on component, as for me it can be moved to the handler

dmitryk-dk (approved):
LGTM!

archef2000:
There still is a problem when unselecting an item in all Editors using the MultiSelect component, as the `values` parameter to the onChange handler is empty.

szibis:
Any ETA of releasing this datasource improvements for builder ??

archef2000:
No as @valyala has not yet answered. The value prediction is completly wrong as the implementation does not match the official docs that I went after and there are some minor problems with some Editors that need fixing.

archef2000:
@kurayama Can confirm correctly working field_values/field_names endpoints.

Loori-R:
@archef2000 could you please clarify if there are any **critical** issues in this PR?
If not, I’d suggest merging it (since it’s already quite large and has been open for a while), and then handling further improvements in separate issues/PRs.

archef2000:
Should a version number be added to the `Operations.tsx` file?

Loori-R:
> Should a version number be added to the `Operations.tsx` file?

Yes, adding a version VictoriaLogs would help track changes.

Loori-R:
> Should I add it in the OperationDefinitions class as attribute or as a comment at the top of the file?

I think adding it as a comment at the top of the file should be sufficient.

archef2000:
@arturminchukov @Loori-R I am now finished with implementing all the operations and fixed all errors I could find.

arturminchukov:
>I get the same problem and traced it to the request select/logsql/stream_field_names just returning Hello from VM data source!

Yeah, you're right! Right now VL datasource doesn't have such endpoints. Created the [issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/403) to add it. So could you wait until I add these endpoints and after it  check if `stream` operation is working?

arturminchukov:
> > Yeah, you're right! Right now VL datasource doesn't have such endpoints. Created the [issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/403) to add it. So could you wait until I add these endpoints and after it check if `stream` operation is working?
> 
> Looks like just need to add
> 
> ```go
> mux.HandleFunc("/select/logsql/streams", ds.VLAPIQuery)
> mux.HandleFunc("/select/logsql/stream_field_names", ds.VLAPIQuery)
> mux.HandleFunc("/select/logsql/stream_field_values", ds.VLAPIQuery)
> ```
> 
> to [here](https://github.com/archef2000/victorialogs-datasource/blob/main/pkg/plugin/datasource.go#L58). Could you add this code to your pr?

UPD. I merged [pr](https://github.com/VictoriaMetrics/victorialogs-datasource/pull/404) with these changes. Could you just make rebase on updated main branch?

archef2000:
No as the OperationList provided by grafana/ui just asks the queryModeller for all operation categories without any context.

szibis:
Quick question how far we are with releasing this version of datasource officially?

archef2000:
The QueryModeller is pretty heavy to create and it would need more than just this change as there are many places where a full QueryModeller is used. Then there is the problem that nearly all operations that are placed result in a query that fails. You first need to fill it with information so that also requires knowledge. It also means limiting advanced users that might create a query out of order. And what happens when I drag a pipe to the start? The error code of vl-select is very informative and should be very easy to understand. There is always a learning curve of a new tool and after the first failed query it should be clear what is allowed.

szibis:
> @szibis in a current state query builder doesn't simplify procedure of building logsql expressions. we've evaluated it internally and despite logsql syntax knowledge some people failed to build successful query during the first attempt to use it and only reverse conversion from raw expression to builder mode helped to understand why it failed. In my opinion builder should not allow to select options that produce absolutely invalid query. For example [here](https://github.com/VictoriaMetrics/victorialogs-datasource/blob/55f7fbe9b0b28cadd15274035f8d74d45d69ab24/src/components/QueryEditor/QueryBuilder/QueryBuilder.tsx#L25) value `query.expr` should impact a subset of categories that is passed to QueryModeller, which are allowed to choose on the next step (don't allow to select pipes, when filter is not set, don't allow setting filter, when pipes are set, etc)
> 
> @szibis Have you tried it already?

I can Try to test it but artifacts expired now from last build

For What I am expecting is to simplify Engineers life. 
Current Datasource with beta builder is not really stable and not working properly not mentioning zero feature only simple filter. 
Yes people can use source query but datasources especially for people moving from another storage to VictoriaLogs should be able to "click-ops" easy simple things not even looking into docs or putting them via source section. 
Source section is for more advance users or users need more advance things to do.
I think datasource builder in first place should focus on log management - filtering eg. metadata, deep dive to find any specific event when looking for root causes or just easy and simple build charts graphs from Logs in Dashboards.

arturminchukov:
After discussing with the team, we decided to wait until the [autocomplete API](https://github.com/VictoriaMetrics/VictoriaLogs/issues/914) is ready. We think that, based on this API, it will be possible to build a query builder that helps users build queries. After it, we will return to the query builder.

archef2000:
Wouldn't it be easier to make that a seperate PR it would mainly be the [‎editorHelper.ts‎](https://github.com/VictoriaMetrics/victorialogs-datasource/pull/329/changes#diff-70f2f5f217a6bc968d97251f37c743253637b77d6b7f1aa239a4130b1552a6e1) file that needs changing.

PR #350, opened 2025-07-30
fix regression implemented in the commit 9880f99

dmitryk-dk:
Fixed regression after the plugin check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## tip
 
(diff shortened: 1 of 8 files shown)


AndrewChubatiuk:
@dmitryk-dk could you port these changes to vm datasource as well?

dmitryk-dk:
> @dmitryk-dk could you port these changes to vm datasource as well?

yes, i will do that today

PR #366, opened 2025-08-24
Remain in inQuotes state for escaped double-quotes

damluk:


dmitryk-dk (commented):
Overall looks good! 
Could you please add test cases with normal stream fields please?
It might be confusing that `_msg` field is present in the `_stream`. 

Could you please update your branch with the latest changes and resolve the conflict

dmitryk-dk (approved):
LGTM!

Issue #347, opened 2025-07-29
Default Log Volume visualization shows all logs as unknown level


ClementGautier:
Hi,

Using the datasource plugin with it's default values result in the Log Volume visualization showing all logs as "unknown" level:

<img width="1399" height="203" alt="Image" src="https://github.com/user-attachments/assets/7826bb0f-2044-4e63-a0f1-a02cb11c33cf" />

If you set even one Log Level Rule then it seems to work well:

<img width="1557" height="149" alt="Image" src="https://github.com/user-attachments/assets/f9cb012b-5e11-4620-82b5-74865f141b9a" />

<img width="1405" height="196" alt="Image" src="https://github.com/user-attachments/assets/6fb23027-0fbf-453e-832f-02484609df68" />

I use version 0.18.1 of the pluging on a Grafana v11.3.0 (d9455ff7db) and a VLCluster version v1.25.1

ClementGautier:
@ExpLog Yes, I have added a transform to fix that:

```
transform/log_level_parser:
  error_mode: ignore
  log_statements:
    - context: log
      statements:
        - set(attributes["level"], ConvertCase(attributes["level"], "lower"))
```

It doesn't fix the issue with the default rules though

PR #370, opened 2025-08-30
fix: level detection logic

Loori-R:
1. Fixed an issue where **level** on the *Logs volume* chart was not detected if no custom rules were set.
2. Fixed an issue where the **level** field was ignored when a custom rule was present — now the existing level is used first, and custom only if it’s missing.
3. Unified the **level** detection logic between the *Logs volume* chart and the logs list (they could differ before).

Related issue: #347


src/configuration/LogLevelRules/utils.ts
```diff
@@ -52,5 +65,29 @@ export const resolveLogLevel = (log: Record<string, any>, rules: LogLevelRule[])
     }
   }
 
-  return LogLevel.unknown;
+  return null;
 }
+
+
+// Auto-detects the log level from level/severity-like keys in labels.
```
dmitryk-dk, line 72: Do we need this commented code?

dmitryk-dk (commented):
Could you please check the CI as well?

arturminchukov (approved):
LGTM!

PR #373, opened 2025-09-04
fix issue with concurrent map writes

dmitryk-dk:
Make a port of the fix related to the issue described [here](https://github.com/VictoriaMetrics/victoriametrics-datasource/issues/363). 

The logs datasource behaves in the same way as the metrics datasource in case to process queries


arturminchukov (approved):
LGTM!

PR #387, opened 2025-09-11
fix: interpolate trace ID for internal datasource links

### LINKED ISSUE #386: Internal datasource links in derived fields do not interpolate field values
When using derived fields with internal datasource links (i.e., links to another Grafana datasource), the field value template `${__value.raw}` is not interpolated. Instead, the literal string `"${__value.raw}"` is passed to the linked datasource, resulting in failed queries.

### Steps to Reproduce

1. Configure VictoriaLogs datasource with a derived field for trace correlation:
```json
{
  "name": "TraceID from trace_id",
  "matcherRegex": "trace_id",
  "matcherType": "label",
  "url": "${__value.raw}",
  "urlDisplayLabel": "View Trace",
  "datasourceUid": "victoriatraces"
}
```
2. Query logs that contain a `trace_id` label (e.g., `trace_id="74bec32e156422caa1a5b0737517b0c5"`)
3. Click on the "View Trace" link that appears next to the trace ID

### Expected Behavior

The link should open the traces datasource with a query for the actual trace ID value:
- Query: `74bec32e156422caa1a5b0737517b0c5`

### Actual Behavior

The link opens the traces datasource with the literal template string:
- Query: `${__value.raw}`

This results in an empty query result since no trace matches the literal string `"${__value.raw}"`.

### System Info

- VictoriaLogs datasource version: 0.20.0
- Grafana version: 12.1.1
- Affected datasource types: any datasource used as internal link target

### Impact
Breaks log-to-trace correlation for any setup using derived fields to link from VictoriaLogs to a tracing datasource

cmacrae:
Previously, internal links to tracing datasources (Jaeger, Zipkin, Tempo, etc.) were incorrectly using the derivedFieldConfig.url value, which is meant for external URL templates. This caused trace links to open with empty queries.

Now internal links always use `${__value.raw}` to properly interpolate the actual trace ID value from the log field, ensuring trace views open with the correct query.

Fixes #386

---

Tested and working on Grafana v12.1.1 :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## tip
 
+* BUGFIX: fix internal datasource links in derived fields to properly interpolate field values instead of passing literal `${__value.raw}` strings. See [#386](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/386).
+
 ## v0.20.0
 
 * FEATURE: upgrade Go builder from Go1.24.2 to Go1.25. See [Go1.25 release notes](https://tip.golang.org/doc/go1.25).
--- a/src/getDerivedFields.ts
+++ b/src/getDerivedFields.ts
@@ -87,9 +87,10 @@ function fieldFromDerivedFieldConfig(derivedFieldConfigs: DerivedFieldConfig[]):
         // Will be filled out later
         title: derivedFieldConfig.urlDisplayLabel || '',
         url: '',
-        // This is hardcoded for Jaeger or Zipkin not way right now to specify datasource specific query object
         internal: {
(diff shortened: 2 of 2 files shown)


cmacrae:
Alright, I figured out what the issue was 🤦 the provisioning method I was using for creating my datasources declaratively wasn't properly configuring the `derivedFields.url` value. Grafana's UI was slightly misleading in displaying the value I was _trying_ to set, but when I checked the datasource's JSON model via the API, I could see that the `derivedFields.url` values were in fact empty... hence why the links were broken when I'd click them

I've fixed the source of the problem in my declarative config and it's now working! 🥳  
So, this fix and the associated issue I opened are in fact not needed. Thanks very much for your help on this @arturminchukov - I'll close this PR and the issue

Issue #390, opened 2025-09-14
Unpredictable behavior when determining interval for range query


aaspiridonov:
I use range Vlogs query in grafana:
`{service="alerts-log"}
labels.severity:="notice"
| stats by (labels.severity) count() as alerts`
<img width="870" height="203" alt="Image" src="https://github.com/user-attachments/assets/1bfb6620-71df-4ee2-b27f-2c45a164a574" />

When specifying an interval for a range query, the result is unpredictable.

**interval=1h [ok]**
first point
<img width="917" height="443" alt="Image" src="https://github.com/user-attachments/assets/3bacb495-8cc6-4932-918e-7890fa773595" />
second point
<img width="917" height="443" alt="Image" src="https://github.com/user-attachments/assets/20ad90b1-8b7d-4fe3-b9b8-7d8da1c562b5" />

**interval=6h [ok]**
first point
<img width="917" height="579" alt="Image" src="https://github.com/user-attachments/assets/88daa129-d970-48b9-9001-4c88a17a9cf3" />
second point
<img width="917" height="579" alt="Image" src="https://github.com/user-attachments/assets/dfa61325-c5cf-4926-9685-0bc16251699e" />

**interval=8h [not ok]**
first point
<img width="917" height="579" alt="Image" src="https://github.com/user-attachments/assets/fb740372-2659-483e-83c7-473d5d5fbc67" />
second point
<img width="917" height="579" alt="Image" src="https://github.com/user-attachments/assets/bfd6e53b-3136-4eba-995b-d8294411c975" />

**interval=12h [ok]**
first point
<img width="917" height="579" alt="Image" src="https://github.com/user-attachments/assets/ea1b8ac7-496a-4b22-bcd5-00e0df601e4d" />
second point
<img width="917" height="579" alt="Image" src="https://github.com/user-attachments/assets/63b8bc62-eda1-4195-903b-f7a20e625c1e" />

**interval=1d [not ok]**
first point
<img width="917" height="579" alt="Image" src="https://github.com/user-attachments/assets/e1c70c1f-3219-4724-831c-c82f387adcdb" />
second point
<img width="917" height="579" alt="Image" src="https://github.com/user-attachments/assets/16e7ac0a-eabe-4860-8f2c-aecfd8118fbd" />

Grafana: v11.6.5
VLogs grafana plugin: v0.20.0
VLogs: v1.33.1

arturminchukov:
[...]

PR #391, opened 2025-09-17
fix unpredictable behavior when determining an interval for a range query q…

arturminchukov:
Related issue: #390 
Fixed unpredictable behavior when determining an interval for a range query query by updating interval comparison logic for `roundInterval` to use `<` instead of `<=` for consistency with [grafana logic](https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/datetime/rangeutil.ts#L383)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * FEATURE: add run in vmui button. The VMUI URL can be configured in the datasource settings. If not specified, the datasource URL with the path `/select/vmui` will be used. See [#369](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/369).
 
 * BUGFIX: fix unpredictable behavior when determining a `Max Data Points` option for a range query. See [#393](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/393).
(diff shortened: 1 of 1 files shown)


arturminchukov:
> It is likely it requires change here
> 
> https://github.com/VictoriaMetrics/victorialogs-datasource/blob/1d98c2fc7f74058f885fac45f3066d9ee6620155/pkg/utils/utils.go#L492
> 
> 
> The `minInterval` shouldn't be rounded, because this is what user sets. 

You're right. 

>The change doesn't look relevant to what https://github.com/VictoriaMetrics/victorialogs-datasource/issues/390 claims.

Also the interval depends on `MaxDataPoints`, so I fixed `roundInterval` function according to the grafana implementation. So now it works as expected - I see the same interval on the graph, which is set in options by grafana.

> Please, also cover this with tests and port tests and changes, if needed, to victoriametrics datasource.

Victoriametrics datasource doesn't have the problem with `min interval`, but it has problem with `MaxDataPoints`, so will port this fix as well.

arturminchukov:
> Would you mind filing this as a separate PR with its own changelog line and reflection in tests? So far, the changelog line seems off to what was actually done.

Yes, I left in this PR only what is related to the `minInterval'. I will create a pr with the `maxDataPoints` fix as a separate pr.

dmitryk-dk (dismissed):
LGTM!

hagen1778 (dismissed):

hagen1778 (approved):

Issue #389, opened 2025-09-14
Fails to parse syntax in /stats_query but succeeds in /stats_query_range


func25:
Hello, I have a query

```
options(concurrency=1) * | collapse_nums prettify | stats count_uniq(_msg) as uq
```

that can run on `/stats_query_range` but not `/stats_query`, see the [grafana playground](https://play-grafana.victoriametrics.com/explore?schemaVersion=1&panes=%7B%226v7%22:%7B%22datasource%22:%22P566C6523BE02F42C%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22victoriametrics-logs-datasource%22,%22uid%22:%22P566C6523BE02F42C%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22options%28concurrency%3D1%29%20%2A%20%7C%20collapse_nums%20prettify%20%7C%20stats%20count_uniq%28_msg%29%20as%20uq%22,%22queryType%22:%22stats%22%7D,%7B%22refId%22:%22B%22,%22datasource%22:%7B%22type%22:%22victoriametrics-logs-datasource%22,%22uid%22:%22P566C6523BE02F42C%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22options%28concurrency%3D1%29%20%2A%20%7C%20collapse_nums%20prettify%20%7C%20stats%20count_uniq%28_msg%29%20as%20uq%22,%22queryType%22:%22statsRange%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1). I checked through [api call](https://play-vmlogs.victoriametrics.com/select/logsql/stats_query?step=10m&start=2025-09-12T03:38:00.910Z&end=2025-09-13T03:38:00.910Z&query=options(concurrency=1)%20*%20|%20collapse_nums%20prettify%20|%20stats%20count_uniq(_msg)%20as%20uq) and it’s ok. I'm not sure if the VL datasource manipulates the syntax and the backend fails to parse it somehow.

dmitryk-dk:
Hi @func25 ! By default, we add `_time` to the `stats` queries to define the correct time range. I think this might be the issue.
```
q.Expr = utils.AddTimeFieldWithRange(q.Expr, q.TimeRange)
```
[...]

PR #392, opened 2025-09-18
fix an issue with incorrect handling `option` field

dmitryk-dk:
Related issue: #389 
For the `/stats` API we need to set `_time` at the beginning of the query. But when we define `option` in the query, it should go first.

Fix an issue with the incorrect handling `option` field in the query for the `/stats` API. By the [documentation]https://docs.victoriametrics.com/victorialogs/logsql/#query-options it should be passed at the beginning of LogsQL query

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * FEATURE: add run in vmui button. The VMUI URL can be configured in the datasource settings. If not specified, the datasource URL with the path `/select/vmui` will be used. See [#369](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/369).
 
+* BUGFIX: fix unpredictable behavior when determining a `Max Data Points` option for a range query. See [#393](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/393).
 * BUGFIX: fix an issue with the incorrect handling `option` field in the query for the `/stats` API. By the [documentation](https://docs.victoriametrics.com/victorialogs/logsql/#query-options) it should be passed at the beginning of LogsQL query. See [#389](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/389).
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 * FEATURE: add run in vmui button. The VMUI URL can be configured in the datasource settings. If not specified, the datasource URL with the path `/select/vmui` will be used. See [#369](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/369).
 
+* BUGFIX: fix an issue with the incorrect handling `option` field in the query for the `/stats` API. By the [documentation]https://docs.victoriametrics.com/victorialogs/logsql/#query-options it should be passed at the beginning of LogsQL query. See [#389](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/389).
```
arturminchukov, line 7:
```suggestion
* BUGFIX: fix an issue with the incorrect handling `option` field in the query for the `/stats` API. By the [documentation](https://docs.victoriametrics.com/victorialogs/logsql/#query-options) it should be passed at the beginning of LogsQL query. See [#389](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/389).
```

pkg/utils/utils.go
```diff
@@ -635,8 +649,8 @@ func hasTimeField(expr string) bool {
 	parts := strings.Split(expr, "|")
 	if len(parts) > 1 {
 		expr = parts[0]
-		return strings.Contains(expr, timeField)
+		return strings.Contains(expr, timeField+":")
```
arturminchukov, line 652: I think we could delete this string. I will be enough to set `expr = parts[0]` 
```suggestion
```

arturminchukov (approved):
LGTM!

arturminchukov (approved):

Issue #330, opened 2025-07-08
rpc error: code = Unknown desc = frame has different field lengths, field 0 is len 59 but field 1 is len 58


lhns:
Hello,

I encountered a bug which causes the grafana query to sometimes fail if it contains contaminated log lines. I could not find out yet why some log lines cause these problems but it seems to happen especially for application errors. Querying via the VictoriaLogs UI works fine. I have attached an export that contains the problematic lines (I cannot identify them though).
In my setup I send docker logs to an otel collector with the fluentd driver and then I forward them to victorialogs as otel logs.
I use Grafana v12.0.2, VictoriaLogs v1.24.0, and VictoriaLogs DataSource v0.18.1.

This is the error, I get in Grafana:
```
logger=context t=2025-07-08T11:19:33.075498738+02:00 level=error msg="Request Completed" method=POST path=/api/ds/query status=500 time_ms=38 duration=38.790644ms size=107 referer="https://grafana.test/d/Pr0sFkGVz/container-logs?editPanel=2&from=now-5m&orgId=1&refresh=5s&timezone=browser&to=now&var-filter=&var-service_prefix=dependabot_worker" handler=/api/ds/query status_source=downstream errorReason=InternalError errorMessageID=plugin.requestFailureError error="client: failed to query data: Failed to query data: rpc error: code = Unknown desc = frame has different field lengths, field 0 is len 59 but field 1 is len 58"
```

[vmui_logs_2025-07-08_091928.json](https://github.com/user-attachments/files/21118930/vmui_logs_2025-07-08_091928.json)

dmitryk-dk:
Hi @lhns ! I am trying to reproduce your error with data you shared, but no luck. Could you please make a request to the VictoriaLogs directly via API with the query from the data source and share the response ?

dmitryk-dk:
> Hi [@dmitryk-dk](https://github.com/dmitryk-dk) that must be it. Maybe its my fault. I got errors in the past that the msg field is not filled on some log lines but I thought I fixed it by adding this config to VictoriaLogs: `-defaultMsgValue=`. I don't really know why the docker fluentd driver sometimes produces log records with an empty msg field.

I made a fix for that situations, so the data will be displayed but the `_msg` field will be empty. It will help to understand the reason of problem, because before the fix the datasource is hidding real problem

dmitryk-dk:
[...]

PR #402, opened 2025-09-28
fix error when frames not equal

dmitryk-dk:
Fixed the corner case for the logs response when `_stream` field is empty and there are no `_msg` field. But previous logs line contains both.
Improved tests, removed the approach when tests hide the error, like 
```
frame has different field lengths, field 0 is len 59 but field 1 is len 58
```

This fix was come from the [comment](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/330#issuecomment-3324201182) in the [issue](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/330)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## tip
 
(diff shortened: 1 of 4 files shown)


arturminchukov (approved):
Looks good! Could you please add changelog?

Issue #379, opened 2025-09-09
log panels only show last logs unless sorted.


pgassmann:
by default, the log panels only show the last part of the log lines.

Our log panels are are configured to show Oldest first. When I scroll down to the bottom of the panel, I see the last logs of the selected timeframe, but the first logs are not from the start of the timerange. Probably the last 5000 (default line limit) before the end of the range.

Now there's a new option to enable infinite scrolling by loading additional results. This works when the panel is configured with Newest first, then scrolling down loads older logs.

Workaround: 

* add `| sort` to query. Then the first logs of the time range are returned, scrolling to the bottom loads more. "Loading newer logs..."

IMO the manual sorting should not be necessary.

PR #418, opened 2025-10-20
379 log panels only show last logs unless sorted

arturminchukov:
Related issue: #379 

1. upgrade grafana packages to version `12.2.0` , because only from the version `12`  `subscribe` function is available, so the sorting automatically applied to the query only from the version 12 of Grafana;
2. add sort pipe `sort by (_time) asc/desc` if logs are sorted by asc/desc direction, so we can get older logs with the limit < count of logs in the selected timerange;
3. fix jest config by setting `grafanaESModules` according to the updated grafana packages.


dmitryk-dk (approved):
LGTM!

PR #429, opened 2025-11-03
Fix grafana testing comments

arturminchukov:
Comments from grafana plugin verification:
> There are a couple of inputs rendered as link values (href properties) that are not sanitized and therefore may be subject to XSS atacks:
> https://github.com/VictoriaMetrics/victorialogs-datasource/blob/v0.21.2/src/configuration/DebugSection.tsx#L88
> https://github.com/VictoriaMetrics/victorialogs-datasource/blob/v0.21.2/src/configuration/DebugSection.tsx#L51-L64
> This [gzip.NewReader](https://github.com/VictoriaMetrics/victorialogs-datasource/blob/main/pkg/plugin/datasource.go#L578) is not closed. Make sure to Close() the Reader at the end of the function.

Fixed: 
1. Urls are sanitized.
2. gzip reader is closed.
3. text filter interpolation for `*`. Removed extra escaping for `*`.

Improved:
1. error message of using variables with regexp
image

### DIFF
--- a/pkg/plugin/datasource.go
+++ b/pkg/plugin/datasource.go
@@ -575,15 +575,14 @@ func (d *Datasource) VLAPIQuery(rw http.ResponseWriter, req *http.Request) {
 	}
 	defer resp.Body.Close()
 
-	reader := io.Reader(resp.Body)
+	reader := resp.Body
 	if resp.Header.Get("Content-Encoding") == "gzip" {
-		gzipReader, err := gzip.NewReader(reader)
+		reader, err = gzip.NewReader(reader)
 		if err != nil {
 			writeError(rw, http.StatusBadRequest, fmt.Errorf("failed to create gzip reader: %w", err))
 			return
 		}
-		defer gzipReader.Close()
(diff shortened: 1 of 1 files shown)


pkg/plugin/datasource.go
```diff
@@ -573,21 +573,24 @@ func (d *Datasource) VLAPIQuery(rw http.ResponseWriter, req *http.Request) {
 			return
 		}
 	}
+	defer resp.Body.Close()
+
 	reader := io.Reader(resp.Body)
 	if resp.Header.Get("Content-Encoding") == "gzip" {
-		reader, err = gzip.NewReader(reader)
+		gzipReader, err := gzip.NewReader(reader)
```
dmitryk-dk: You do not need to reassign reader; you can just close it on line 584
it would be one line change

arturminchukov: fixed it

makasim:
Follow up PR on https://github.com/VictoriaMetrics/victorialogs-datasource/pull/412

Issue #338, opened 2025-07-16
Add Syntax highlighting and AutoCompletion to the LogsQL editor


sbordeyne:
Currently, it's quite hard to use the victorialogs datasource in grafana for a few reasons. It's hard to discover what kinds of queries you can write, and it's hard to figure out which labels are available, which operations can be piped together and so on.

In order to alleviate those issues, having auto completions and syntax highlighting might ease the pains above.

For reference, here is the MonacoQueryField reference for the loki datasource: https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/loki/components/monaco-query-field/MonacoQueryField.tsx#L104

As you can see, it has hooks for autocompletion, and hooks for custom syntax highlighting in it. As far as usage goes, it would make a world of difference to have common pipe operations proposed after a query and a pipe character, or have labels autocompleted, or encourage the use of streams in the query.

Loki's datasource also includes a Label Browser as well as a "Query Kickstarter" which gives out common queries you can use.

hagen1778:
@Loori-R @arturminchukov could you pls take a look?

chenlujjj:
+1, that's a very desired feature for users trying to move from Loki, Splunk, etc.

PR #448, opened 2025-11-15
WIP: Add query autocomplete in code builder

sbordeyne:
Related to issue #338 

I started implementation of syntax highlighting and autocompletion in the victorialogs datasource plugin.

I took the implementation from Loki's datasource as the base, and am working on gradually modifying the implementation to fit VictoriaLog's LogsQL instead of Loki's LogQL, the two are still quite similar, thus it's a good starting point.

This PR is currently a work in progress, and not ready to merge

Currently supported:

- Stream labels and value autocomplete
- Basic unpack_$parser pipe

---

**Note:** I added a few services in `compose.yaml` to add logs to VictoriaLogs for manual testing during development, if needed I can easily revert the changes from that commit.


arturminchukov:
@sbordeyne thanks for pr. Currently, there is a plan to implement the autocompletion feature on the part of VictoriaLogs. This will enable us to have the same auto-completion mechanism in both the plugin and the Victoria Logs UI, as well as support the newly added LogsQL functions. We should wait for this feature to avoid unnecessary effort costs. Will duplicate this message to the related issue to update the status.

sbordeyne:
> Not only because of the duplication of code, but also because the LogsQL parser has already been implemented on the backend, which is necessary to provide suggestions correctly. Parser could be changed in the future(e.g., adding new functions), so it will be easier and better just to maintain one place in code.

A parser is also required to provide syntax highlighting, does that mean that syntax highlighting won't come to the datasource ? I can't imagine trying to write a query and have the colors come in after a 100ms delay.

> Currently, you need to do api calls to get the fieldnames and fieldvalues

True but that can be cached on the frontend pretty easily, which is what the loki datasource does already.

Regardless, I won't challenge the decision to do autocompletion on the backend further, though what would be the ETA on that feature coming to the datasource ? Are there endpoints done already to provide completions for a partial query ? I could do the implementation on the datasource side if that would help

arturminchukov:
The autocomplete feature could be implemented using the [autocomplete API](https://github.com/VictoriaMetrics/VictoriaLogs/issues/914). But need to wait for this API. Right now, I don't have an ETA for this.

Issue #465, opened 2025-11-23
Log Level Rules are not working as expected: level is not shown as label in the Log Panel


h4fn1x:
### Describe the bug

Hi,

I defined some Log Level Rules in the Datasource Options (see screenshot), but they are not working as I would expect.

The level is not shown as label in the "Logs" panel, as it is the case with a Loki source (again, see screenshots below). The `detected_level` field shows `unknown`, with Loki it shows the detected level.

But somehow the level must have been detected, because it's shown in the "Logs volume" panel.

I don't know if this is a bug or some misconfiguration by my side...

### Rules:
<img width="3396" height="398" alt="Image" src="https://github.com/user-attachments/assets/a29eebe0-b123-48f5-84b5-197b7b131e8f" />

### Output with VictoriaLogs:
<img width="3444" height="1986" alt="Image" src="https://github.com/user-attachments/assets/097f7f5e-e8fd-467e-93d9-d485f91b6dbb" />

### Output with Loki:
<img width="3434" height="2132" alt="Image" src="https://github.com/user-attachments/assets/b070771b-3259-4f7b-961e-94582e726de4" />

### To Reproduce

See description.

### Versions of VictoriaLogs datasource and VictoriaLogs backend

Datasource: v0.22.0
VictoriaLogs: v1.38.0
Grafana: v12.3.0

PR #470, opened 2025-11-26
fix an issue where `_msg` was incorrectly parsed to `detected_level` label according to `Log Level rules`

arturminchukov:
Related issue: #465 
Fixed an issue where `_msg` was incorrectly parsed to `detected_level` label according to `Log Level rules`. Added `_msg` from `Lines` field to `labels` object during the detection of the level.

### DIFF
--- a/src/backendResultTransformer.ts
+++ b/src/backendResultTransformer.ts
@@ -1,13 +1,13 @@
 import {
   DataFrame,
+  DataFrameType,
   DataQueryError,
   DataQueryRequest,
   DataQueryResponse,
   Field,
   FieldType,
   isDataFrame,
-  QueryResultMeta,
-  DataFrameType
+  QueryResultMeta
 } from '@grafana/data';
 
 import { LogLevelRule } from "./configuration/LogLevelRules/types";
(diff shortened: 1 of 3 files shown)


src/backendResultTransformer.ts
```diff
@@ -52,11 +53,14 @@ function setFrameMeta(frame: DataFrame, meta: QueryResultMeta): DataFrame {
 
 function addLevelField(frame: DataFrame, rules: LogLevelRule[]): DataFrame {
   const rows = frame.length ?? frame.fields[0]?.values.length ?? 0;
+  const lineField = frame.fields.find(f => f.name === FrameField.Line);
   const labelsField = frame.fields.find(f => f.name === FrameField.Labels);
 
   const levelValues = Array.from({ length: rows }, (_, idx) => {
     const labels = (labelsField?.values[idx] ?? {}) as Record<string, any>;
-    return extractLevelFromLabels(labels, rules);
+    const msg= lineField?.values[idx] ?? '';
```
Loori-R:
```suggestion
    const msg = lineField?.values[idx] ?? '';
```

Issue #475, opened 2025-12-02
Add select of tenants in datasource configuration


arturminchukov:
Change the input to select in the multitenancy section. To get a list of tenants, use the [/select/tenant_ids endpoint](https://docs.victoriametrics.com/victorialogs/querying/#querying-tenants) starting from [v1.38.0 release](https://docs.victoriametrics.com/victoriametrics/changelog/).
<img width="569" height="187" alt="Image" src="https://github.com/user-attachments/assets/3d8dcb21-8e57-4210-9d5b-88d73a22ff4d" />

arturminchukov:
PR has been merged and will be included in the next release.

PR #487, opened 2025-12-04
enable tenants selector

dmitryk-dk:
Enable the tenants selector into the datasource settings. 
Screenshot 2025-12-05 at 13 26 07
If the backend doesn't support the endpoint the previous inputs will be used
image

Related issue: https://github.com/VictoriaMetrics/victorialogs-datasource/issues/475

### DIFF
--- a/src/configuration/TenantSettings.tsx
+++ b/src/configuration/TenantSettings.tsx
@@ -22,6 +22,7 @@ const documentationLink = (
 export const TenantSettings = (props: PropsConfigEditor) => {
   const { options, onOptionsChange } = props;
   const multitenancyHeaders = options.jsonData?.multitenancyHeaders;
+  const isReadOnly = options.readOnly;
 
   const [tenants, setTenants] = useState<ComboboxOption[]>([]);
   const [isLoading, setIsLoading] = useState(true);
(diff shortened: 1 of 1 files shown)


Loori-R:
Do I understand correctly that the user can select `ProjectID` and `AccountID` independently from each other?
Shouldn’t they be selected as a pair (i.e., only valid `ProjectID:AccountID` combinations)?

dmitryk-dk:
> Do I understand correctly that the user can select `ProjectID` and `AccountID` independently from each other? Shouldn’t they be selected as a pair (i.e., only valid `ProjectID:AccountID` combinations)?

yep. But if we have 
```
{[accountid:5, projectid:1], [accountid: 5, projectid:2], [accountid: 5, projectid:3]}
```
which to choose (first)?
If yes i will update the logic

Loori-R:
I think the simplest approach is to have a single selector with values in the `AccountID:ProjectID` format (same as in vmui and the format returned by VictoriaMetrics). This way the user always selects a valid pair.

An alternative would be to keep two fields and filter `ProjectID` options based on the selected `AccountID`, but this seems less convenient - the user would first have to change the `AccountID` in order to access a different `ProjectID`.

dmitryk-dk:
> I think the simplest approach is to have a single selector with values in the `AccountID:ProjectID` format (same as in vmui and the format returned by VictoriaMetrics). This way the user always selects a valid pair.
> 
> An alternative would be to keep two fields and filter `ProjectID` options based on the selected `AccountID`, but this seems less convenient - the user would first have to change the `AccountID` in order to access a different `ProjectID`.

Oki doki, will update to one selector :)

PR #596, opened 2026-02-26
feat: add opt-in dataplane format

github-vincent-miszczak:
Add an opt-in setting and the corresponding UI toggle to enable logs [dataplane](https://grafana.com/developers/dataplane/logs) format.

Using this format allows Grafana correlations to work correctly with label values.
Without it, it's not possible to craft links that use label values when using Grafana correlations.

It's opt-in because the fields are renamed, which may break client-side transformations.

---
## Summary by cubic
Adds an opt-in setting to output logs in Grafana dataplane format, enabling automatic `${labelKey}` variables in correlations. Default is off to avoid breaking transformations that reference `Time`/`Line`.

- **New Features**
  - New “Use dataplane format” toggle in datasource settings; provisioning key: `jsonData.useDataplaneFormat`.
  - When enabled: rename `Time`→`timestamp` and `Line`→`body`, and set frame meta to `DataFrameType.LogLines` with `typeVersion [0,0]` so `${labelKey}` works in correlations.
  - Updated README, CHANGELOG, and tests to cover the new format.

<sup>Written for commit e49b2eae59398482dc282db460cd625832b57465. Summary will update on new commits.</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * FEATURE: add a `:` (word filter) operator to Log Level Rules, matching a field value by whole word like LogsQL (e.g. `error` matches the word `error`, not `terror`). See [#611](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/611).
 * FEATURE: the visual query builder now shows context-aware keyboard hints at the bottom of its dropdowns, so the available shortcuts (select, navigate, confirm, run) are visible while you build a query. See [pr #683](https://github.com/VictoriaMetrics/victorialogs-datasource/pull/683).
+* FEATURE: add opt-in Grafana dataplane log format support. When enabled in datasource settings, log frames use standard field names (`timestamp`, `body`) and `DataFrameType.LogLines` metadata, enabling automatic `${labelKey}` variables in correlations.
 
 ## v0.29.0
 
--- a/src/README.md
+++ b/src/README.md
@@ -46,6 +46,8 @@ datasources:
     url: http://victorialogs:9428
     isDefault: true
     jsonData:
+      # Enable Grafana dataplane log format for correlations support
+      #useDataplaneFormat: true
       # Multitenancy settings, see https://docs.victoriametrics.com/victorialogs/#multitenancy
       # to use the multitenancy, uncomment lines below: AccountID and ProjectID
       multitenancyHeaders:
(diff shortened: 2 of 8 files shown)


arturminchukov:
@github-vincent-miszczak Thanks for contributing! I didn't understand how it could help with the correlations. I'm just tried to set up a correlation but instead of `Line` field we just get the `body` field. I would be glad, if could share me a guide with setting up a correlation with a data plane format.

github-vincent-miszczak:
Here is an example using the dataplane format without any transformation.
The link is correctly dynamically generated with our field value.

Label values get automagically promoted as top level variables.
Without this format, there is no syntax to access label values. At best we can workaround using transformation/regex, but that's not ideal.

image

image

image

arturminchukov:
@github-vincent-miszczak Could you please sign your commits? [Here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), you can find a guide on how to do it.

arturminchukov:
@github-vincent-miszczak 
Right now, to use Grafana correlations (like linking from logs to traces), you have to enable the dataplane format. The problem is that it renames Time to timestamp and Line to body, which can break existing dashboards and panels that rely on those field names.

 What I suggest: the plugin reads the correlation configs that the user has set up in Grafana, looks at which label field each correlation needs, pulls that value out of the log labels, and creates a clickable field for it — all at query time, automatically.

 So the user just creates a correlation in the Grafana UI (say, logs → Tempo with trace_id), and it works. No need to enable any toggles, no field renaming, no risk of breaking anything.

 To get the correlation configs at runtime, we use `getCorrelationsService()` from `@grafana/runtime` — specifically `getCorrelationsBySourceUIDs([datasourceUid])` which returns all correlations where our datasource is
 the source. Each correlation has a `config.field `telling us which label to extract.
What do you think?

github-vincent-miszczak:
Hi, thanks for taking the time look at this.

Yes, the field rename is annoying, but the opt-in toggle keeps it safe. It defaults to off, so no existing dashboard breaks. Users who need correlations flip it on knowingly.

It's the Grafana standard. The [logs dataplane contract](https://grafana.com/developers/dataplane/logs) is the documented way for log datasources to expose label values to features like correlations, and Grafana core already special-cases it in many places (logsFrame.ts, LogDetails.tsx, LogLineDetailsComponent.tsx, the logs-table extractLogsFieldsTransform, panel suggestions, etc.). Building a plugin-side workaround duplicates what the platform already provides natively.

Correlations aren't the only consumer. Label promotion happens in getFieldLinksForExplore (https://github.com/grafana/grafana/blob/main/public/app/features/explore/utils/links.ts)): when frame.meta.type === DataFrameType.LogLines, every label is added to scopedVars as a flat ${labelKey}. That same path serves both correlation links and regular data links / derived fields — not correlation-specific. 

Future-proofing label metadata. The dataplane spec also defines an optional labelTypes field (Grafana 12.4+) that lets us distinguish stream fields from parsed/computed fields in Log Details via DataSourceWithLogsLabelTypesSupport. That's the standard path for surfacing VictoriaLogs' stream-vs-log-field distinction in the UI.

Less plugin code, less drift. The dataplane path is just a field rename + frame meta — minimal surface area.

arturminchukov:
@github-vincent-miszczak could you please rebase on main?

Issue #599, opened 2026-03-02
In Health endpoint make a request to query


arturminchukov:
Change the health endpoint in the DS backend. Instead of making a request to the `health` endpoint, make a request to the `query` endpoint with the `query=*`. It will help to check if there is a real VL backend.

PR #627, opened 2026-04-22
Otel preset

arturminchukov:
Related issue: #599

### Describe Your Changes

1. Changed `/health` check. Now send a real request to logs with `query=*` to ensure that the endpoint is correct.(Issue: #599)
2. Added OpenTelemetry preset for automatic OTel field and severity detection Queries recent logs to detect field naming format (snake_case vs camelCase) and severity field with its value case (UPPER/lower/mixed), then generates derived fields for trace ID and log level rules for OTel severity — without manual configuration.

### Checklist

---
## Summary by cubic
Adds an OpenTelemetry preset that auto-detects OTel trace/severity fields and generates derived fields and log level rules. Save & Test now runs a real log query for a meaningful health check. Related to #599.

- **New Features**
  - OpenTelemetry preset in datasource settings that detects `trace_id`/`traceId`/`TraceId` and a severity field among `severity_number`/`severityNumber`/`SeverityNumber`/`severity_text`/`SeverityText`/`severity`/`Severity`; classifies severity as `number` or `string`.
  - Auto-generates a trace ID derived field (label matcher, links to an optional traces datasource) and OTel-based log level rules: case-insensitive equals for text, or 6 regex ranges for severity numbers; preserves existing user fields/rules. UI lets you re-run detection, change the severity field, and pick an optional traces datasource.

- **Bug Fixes**
  - Replace the `/health` ping with a real instant query to `/select/logsql/query` (`query=*`, `limit=1`, last 5m) so Save & Test catches bad tenants, auth, or storage; forwards multitenancy and custom headers and respects the configured HTTP method.

<sup>Written for commit 7203327b376f511704eec857394c2f3e6274ede5. Summary will update on new commits.</sup>


pkg/plugin/datasource.go
```diff
@@ -486,18 +486,44 @@ func checkAlertingRequest(headers map[string]string) (bool, error) {
...
+		return newHealthCheckErrorf("failed to parse datasource URL: %s", err), nil
+	}
+	u.Path = path.Join(u.Path, instantQueryPath)
+
+	now := time.Now()
+	values := u.Query()
+	values.Set("query", "*")
+	values.Set("limit", "1")
+	values.Set("start", strconv.FormatInt(now.Add(-24*time.Hour).Unix(), 10))
+	values.Set("end", strconv.FormatInt(now.Unix(), 10))
```
hagen1778: I think that's ok to skip end param, as I don't see how it helps here. If end is omitted, it is automatically set to now() by VL

pkg/plugin/datasource.go
```diff
@@ -486,18 +486,44 @@ func checkAlertingRequest(headers map[string]string) (bool, error) {
...
+	if err != nil {
+		return newHealthCheckErrorf("failed to parse datasource URL: %s", err), nil
+	}
+	u.Path = path.Join(u.Path, instantQueryPath)
+
+	now := time.Now()
+	values := u.Query()
+	values.Set("query", "*")
+	values.Set("limit", "1")
+	values.Set("start", strconv.FormatInt(now.Add(-24*time.Hour).Unix(), 10))
```
hagen1778:
```suggestion
	values.Set("start", "-5m")
```

See https://docs.victoriametrics.com/victoriametrics/single-server-victoriametrics/#timestamp-formats

pkg/plugin/datasource.go
```diff
@@ -486,18 +486,44 @@ func checkAlertingRequest(headers map[string]string) (bool, error) {
...
 		res.Status = backend.HealthStatusError
 		res.Message = "Error getting datasource instance"
 		d.logger.Error("Error getting datasource instance", "err", err)
 		return res, nil
 	}
-	r, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s%s", strings.TrimRight(di.settings.URL, "/"), health), nil)
+	return d.checkHealthWithInstance(ctx, di)
+}
+
+func (d *Datasource) checkHealthWithInstance(ctx context.Context, di *DatasourceInstance) (*backend.CheckHealthResult, error) {
```
hagen1778: should this function belong to `d *Datasource`? I don't see `d` to be used insinde.
Could it be just the following?
```
func checkInstanceHealth(ctx context.Context, di *DatasourceInstance) (*backend.CheckHealthResult, error) 
```

pkg/plugin/datasource.go
```diff
@@ -486,18 +486,44 @@ func checkAlertingRequest(headers map[string]string) (bool, error) {
...
+	u.RawQuery = values.Encode()
+
+	method := di.grafanaSettings.HTTPMethod
+	if method == "" {
+		method = http.MethodGet
+	}
+
+	r, err := http.NewRequestWithContext(ctx, method, u.String(), nil)
 	if err != nil {
 		return newHealthCheckErrorf("could not create request"), nil
```
hagen1778: please add the actual error

pkg/plugin/datasource.go
```diff
@@ -486,18 +486,44 @@ func checkAlertingRequest(headers map[string]string) (bool, error) {
...
 	}
+	// CustomHeaders includes multitenancy headers (AccountID, ProjectID) and user-defined headers.
+	r.Header = di.grafanaSettings.CustomHeaders.Clone()
+
 	resp, err := di.httpClient.Do(r)
 	if err != nil {
 		return newHealthCheckErrorf("request error"), nil
```
hagen1778: it would be nice to add actual errors here and in other places too

hagen1778 (changes requested):

src/components/QueryEditor/QueryHints/hints/useLevelQueryHintSection.ts
```diff
@@ -29,9 +28,13 @@ export const useLevelQueryHintSection = (levelRules: LogLevelRule[]): QueryHintS
 
     const hints = Object
       .entries(levelFilters)
-      .map(([ruleLevel, rule]): LevelHint => {
+      .map(([ruleLevel, rules]): LevelHint => {
         const levelKey = ruleLevel as UniqLogLevelKeys;
-        const queryExprByRules = rule.map(r => `${r.field}:${OperatorLabels[r.operator]}"${r.value}"`);
+        const buildQueryExpr = (rule: LogLevelRule) => {
```
dmitryk-dk: nit. can we make it as separate handler instead if initialise it on every map iteation?

src/configuration/LogLevelRules/const.ts
```diff
@@ -19,12 +19,22 @@ export const LOG_OPERATOR_OPTIONS: LogOperatorOption[] = [
 
 export const OperatorLabels: Record<LogLevelRuleType, string> = {
   [LogLevelRuleType.Equals]: '=',
+  [LogLevelRuleType.CaseInsensitiveEquals]: '~',
```
dmitryk-dk, line 22: which a difference between `[LogLevelRuleType.Regex]: '~',` ?

src/configuration/OpenTelemetryPreset/DetectedFormatCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+  onStartChangingSeverity,
+  onCancelChangingSeverity,
+}: DetectedFormatCardProps) => (
+  <Card>
+    <Stack direction='column' gap={1}>
+      <Text variant='bodySmall'><strong>Detected format</strong></Text>
+      <Text variant='bodySmall'>
+        <strong>Trace field</strong>
+        <Tooltip content={<TraceFieldTooltip />} placement='top'>
+          <Icon name='info-circle' size='sm' style={{ marginLeft: 4, marginRight: 4, cursor: 'help' }} />
```
dmitryk-dk: do we need style defined like object or we can use grafanaStyles?

src/configuration/OpenTelemetryPreset/useOtelPreset.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  // Re-run health check when the datasource is saved (id/version change after "Save & test").
+  useEffect(() => {
+    detectionTriggeredRef.current = false;
+
+    const check = async () => {
+      try {
+        const ds = await getDataSourceSrv().get(options.uid) as VictoriaLogsDatasource;
+        const health = await ds.callHealthCheck();
+        if (health.status === HealthStatus.OK) {
```
dmitryk-dk, line 62: Maybe we can use the same const `HealthStatus.OK` everywhere intead of strings `ok`, `errror`, just a proposition

src/configuration/LogLevelRules/const.ts
arturminchukov: It uses the different logsql function: `contains_common_case`, but for short displaying of this function, I use `~`.

src/components/QueryEditor/QueryHints/hints/useLevelQueryHintSection.ts
arturminchukov: makes sense, thanks!

hagen1778 (approved):
LGTM

dmitryk-dk (approved):
LGTM!

PR #638, opened 2026-05-11
fix(query interpolation): prevent the wildcard `*` value of a multi-value template variable from being wrapped in double quotes inside the `in(...)` operator during query interpolation

arturminchukov:
### Describe Your Changes

Prevent the wildcard `*` value of a multi-value template variable from being wrapped in double quotes inside the `in(...)` operator during query interpolation. 

Example: 
```
tenant:in($tenant)
| query_hash:$query_hash 
| type:="range"
| count()
```
with the textbox variable `query_hash` and value `*` was incorrectly interpolated to

```
tenant:in("0")
| query_hash:in("*")
| type:="range"
| count()
```
.

Now the interpolated query is:
```
tenant:in("0")
| query_hash:in(*)
| type:="range"
| count()
```
### Checklist

---
## Summary by cubic
Fix query interpolation so the wildcard `*` inside `in(...)` is not quoted for template variables (including textbox). This now generates `in(*)` instead of `in("*")`, restoring correct wildcard behavior.

- **Bug Fixes**
  - Run multi-value replacement immediately after template replacement to keep `*` unquoted in `in(...)`.
  - Added a regression test and changelog entry.

<sup>Written for commit 745d7467b923b892116eb74538a79e00d3a6c1c9. Summary will update on new commits.</sup>

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -364,9 +364,10 @@ export class VictoriaLogsDatasource
     const variableNamesList = this.templateSrv.getVariables().map(v => v.name);
     expr = doubleQuoteRegExp(expr, variableNamesList);
     expr = this.templateSrv.replace(expr, scopedVars, this.interpolateQueryExpr);
+    expr = this.replaceMultiVariables(expr);
     expr = correctRegExpValueAll(expr);
     expr = correctMultiExactOperatorValueAll(expr);
-    return this.replaceMultiVariables(expr);
+    return expr;
   }
 
   private replaceMultiVariables(input: string): string {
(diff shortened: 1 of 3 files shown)


Loori-R (approved):
LGTM!

src/datasource.ts
```diff
@@ -364,9 +364,10 @@ export class VictoriaLogsDatasource
     const variableNamesList = this.templateSrv.getVariables().map(v => v.name);
     expr = doubleQuoteRegExp(expr, variableNamesList);
     expr = this.templateSrv.replace(expr, scopedVars, this.interpolateQueryExpr);
+    expr = this.replaceMultiVariables(expr);
```
hagen1778, line 367: @arturminchukov could you please cover this case in tests?

hagen1778 (approved):

Issue #563, opened 2026-02-11
Do not sort fields received from VictoriaLogs at plugin side


valyala:
## The issue

Previously VictoriaLogs didn't give any guarantees regarding the order of fields it returns in query responses, so response fields could be returned in an arbitrary order. If the response contains many fields, it would be great sorting them for easier investigation by humans. That's why it was recommended to sort the received fields by name by clients, which query VictoriaLogs, such as [built-in web UI for VictoriaLogs](https://docs.victoriametrics.com/victorialogs/querying/#web-ui) and the Grafana plugin for VictoriaLogs.

The commit https://github.com/VictoriaMetrics/VictoriaLogs/commit/db81556d7316e464f69926e2d85894067005fc66 implemented such a guarantee. Now VictoriaLogs consistently returns fields sorted by name for all the queries except those queries, which end with a [pipe](https://docs.victoriametrics.com/victorialogs/logsql/#pipes), which preserves the order of output fields. The most frequently used pipes, which preserve the order of the returned fields, are [`fields` pipe](https://docs.victoriametrics.com/victorialogs/logsql/#fields-pipe) and [`stats` pipe](https://docs.victoriametrics.com/victorialogs/logsql/#stats-pipe). If the query ends with such pipes, then VictoriaLogs preserves the order of the requested fields in the query response.

This removes the need to sort the returned fields at client side.

## The solution

To remove all the sorting of the returned log fields at the VictoriaLogs plugin for Grafana side.

See https://github.com/VictoriaMetrics/VictoriaLogs/issues/1011#issuecomment-3884341383 for more details.

Issue #665, opened 2026-06-03
Show stream labels in context modal


arturminchukov:
### Problem
Today the context query uses only `_stream_id` and shows **no UI** — users
can't broaden the search beyond the exact stream:

### Request

Show the row's stream labels as toggleable chips in the context modal
(like Loki's getLogRowContextUi). Disabling a label widens the search;
the default (unchanged) keeps the current _stream_id behavior.

mitsurugi-dev:
@arturminchukov Good afternoon, please tell me, I didn't understand much of this comment.
<img width="699" height="94" alt="Image" src="https://github.com/user-attachments/assets/b6f86f31-023b-4e0a-86e8-1e72e4ead563" />

I have Fields like this, weren't they supposed to be like Labels in Loki?
<img width="183" height="295" alt="Image" src="https://github.com/user-attachments/assets/18e32acd-b9aa-4fda-8c16-ea4763170506" />

mitsurugi-dev:
@arturminchukov Hi, can you please tell me about this issue above?

arturminchukov:
[...]

PR #666, opened 2026-06-04
feat(logs): preserve field order from VictoriaLogs response

arturminchukov:
Related issue: #563 

### Describe Your Changes

- Serialize log fields via fastjson Del + MarshalTo instead of a Go map, so the labels JSON keeps VL's field order instead of an alphabetical sort
- Stop parsing _stream into fields (VL already expands stream labels as top-level fields); keep the raw _stream string for the stable log id
- Remove the now-unused orderedLabels type and pkg/utils/stream_fields_parser
- Update test fixtures to real VL top-level expansion; drop invalid_stream case

### Checklist

---
## Summary by cubic
Preserves VictoriaLogs field order in labels JSON and stops parsing `_stream`. Builds stable log IDs from raw `_time`, `_msg`, and `_stream_id` bytes (keyed on `_stream_id`) to match VL responses and improve stability (#563).

- **Refactors**
  - Serialize labels with `fastjson` (`Del` unwanted fields, then `MarshalTo`) so log fields keep VL order; `_stream`/`_stream_id` are kept as-is.
  - Split responses and tests into `response_logs.go`, `response_stats.go`, `response_hits.go`; keep shared types/error parsers/`labelsToJSON` in `response.go`.
  - Remove `pkg/utils/stream_fields_parser`; update fixtures and drop `pkg/plugin/test-data/invalid_stream`.

<sup>Written for commit e0db15425d0e92ddfc838af7aa069b36b5826e37. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VictoriaMetrics/victorialogs-datasource/pull/666?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pkg/plugin/response.go
```diff
@@ -113,42 +114,21 @@ func parseInstantResponse(reader io.Reader) backend.DataResponse {
...
-			stream := value.GetStringBytes(streamField)
-			rowStream = string(stream)
-			stf, err := utils.ParseStreamFields(rowStream)
-			if err != nil {
-				return newResponseError(fmt.Errorf("%s", err), backend.StatusInternal)
-			}
-			for _, field := range stf {
-				labels[field.Label] = field.Value
-			}
-			if !value.Exists(messageField) && len(stf) >= 0 {
```
arturminchukov, line 127: Removed stream parsing, cause stream labels always exist in the log as separate labels, so we don't need to parse them manually.

dmitryk-dk (changes requested):
As for me we shouldn't change tests files because those are real cases which was fixed via the real users issue.
If we just need to leave `_stream` fields we can update only this logic and tests related to it

pkg/plugin/response.go
```diff
@@ -113,42 +114,21 @@ func parseInstantResponse(reader io.Reader) backend.DataResponse {
...
+			// pad lineField on rows without _msg to keep frame fields equal length
+			if !value.Exists(messageField) {
 				lineField.Append("")
 			}
 		}
 
-		obj, err := value.Object()
-		if err != nil {
-			return newResponseError(fmt.Errorf("error get object from decoded response: %s", err), backend.StatusInternal)
+		if value.Type() != fastjson.TypeObject {
```
hagen1778: Should we check it right after `parser.ParseBytes(b)` was called above? That should make code more clear

pkg/plugin/response.go
```diff
@@ -113,42 +114,21 @@ func parseInstantResponse(reader io.Reader) backend.DataResponse {
...
-			labels[fieldName] = value
-		})
 
-		d, err := labelsToJSON(labels)
-		if err != nil {
-			return newResponseError(err, backend.StatusInternal)
-		}
-		labelsField.Append(d)
+		value.Del(timeField)
+		value.Del(messageField)
```
hagen1778: why not deleting streamField too?
Would it be better to del these fields only if they're exist (in their `if` clause above)?

pkg/plugin/response.go
```diff
@@ -113,42 +114,21 @@ func parseInstantResponse(reader io.Reader) backend.DataResponse {
...
+		labelsField.Append(json.RawMessage(value.MarshalTo(nil)))
 		idField.Append(buildLogID(rowTime, rowMsg, rowStream))
```
hagen1778: buildLogID uses all 3 params as byte slices - would it make sense to pass byte slice instead of strings?  That would reduce allocations and should make func clearer

hagen1778 (changes requested):
It would be nice to refactor datasource_test.go in future. I think it is too verbose and repetetive - that's the reason why it got so many changes in this PR

pkg/plugin/response.go
arturminchukov: >why not deleting streamField too?

I left the `_stream` field to show stream labels in the context modal. Without this field, we cannot detect which labels belong to the stream. It will be done in a separate PR - see https://github.com/VictoriaMetrics/victorialogs-datasource/issues/665.

hagen1778: It would be nice to have a comment explaining why you did this

arturminchukov:
1st commit: 
 - Split the `response.go` into `response_logs.go`, `response_stats.go`, and `response_hist.go` with their tests without any changes in functionality.

2-nd commit:
 - removed parsing `_stream` field.
 - changed function `buildLogID` - now it requires byte arguments and changed `_stream` field to `_stream_id` field.
 - fixed tests according to missing `parse stream` functionality (just removed unnecessary test, and fixed those tests that expected parsing `_stream` field)


@dmitryk-dk @hagen1778 Could you please review it again?

pkg/plugin/datasource_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/VictoriaMetrics/victorialogs-datasource/pkg/utils"
```
hagen1778: prefer the following order of imports:
1. Standard lib
2. External dependencies
3. Internal dependencies

hagen1778 (approved):
LGTM

dmitryk-dk (approved):
Overall looks good!
One thing what i do not like is that we removed from the tests the real test case
but if you sure it is expected than ok
```goalng
_, err := w.Write([]byte(`{"_time":"2024-02-20", "_stream":"{application=\"logs-benchmark-Apache.log-1708437847\",hostname=}"}`))
			if err != nil {
				t.Fatalf("error write reposne: %s", err)
			}
```

arturminchukov:
> Overall looks good! One thing what i do not like is that we removed from the tests the real test case but if you sure it is expected than ok
> 
> ```
> _, err := w.Write([]byte(`{"_time":"2024-02-20", "_stream":"{application=\"logs-benchmark-Apache.log-1708437847\",hostname=}"}`))
> 			if err != nil {
> 				t.Fatalf("error write reposne: %s", err)
> 			}
> ```

I removed this test case, cause I removed the parsing `_stream` field, so this test case is no longer relevant.

hagen1778:
Removing `_stream` field from response parsing was a mistake and will be reverted in https://github.com/VictoriaMetrics/victorialogs-datasource/pull/685
The use case where this matters are queries that rely on _stream field in response:
- `* | field_values _stream`
- `* | fields _stream`

If _stream field will be dropped from response - these queries won't return useful info.

PR #672, opened 2026-06-17
pkg/plugin: pad missing _time rows in instant log frames

immanuwell:
Related issue: n/a

### Describe Your Changes

Fix an edge case in instant log parsing.

If VictoriaLogs returns mixed rows where one row has `_time` and another one does not, the backend built a malformed frame with uneven field lengths, so Grafana just bails out.

This patch pads missing `_time` per row with the same fallback time already used when `_time` is missing for the whole response. Small fix, no extra behavior changes.

Repro:
1. Add this fixture:
   `{"_time":"2024-02-20T14:04:27Z","_msg":"123","_stream":"{app=\"test\"}"}`
   `{"_msg":"456","_stream":"{app=\"test\"}"}`
2. Run `go test ./pkg/plugin -run Test_parseInstantResponse -count=1`
3. Before this patch it fails with `frame has different field lengths`
4. With this patch it passes

Why this is real:
VictoriaLogs docs say logs without `_time` are ingested with the current timestamp, so mixed rows are totally possible in practice, not some weird unreachable corner.

### Checklist

---
## Summary by cubic
Pads missing `_time` values per row in instant log responses to prevent malformed frames and Grafana errors when rows mix with and without `_time`.

- **Bug Fixes**
  - In `parseInstantResponse`, use a single `fallbackTime` from `nowFunc()` to fill missing per-row `_time`, and reuse it when the whole response lacks `_time`.
  - Added a test and fixture for mixed rows to ensure equal field lengths and valid frames.
  - Updated changelog entry to document the fix.

<sup>Written for commit c63483439e426e17472aa6ce71c225b6102aa392. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VictoriaMetrics/victorialogs-datasource/pull/672?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hagen1778:
Thanks for the PR @immanuwell !
Could you please provide example when VictoriaLogs can return a response with entries with/without _time field? I just want to understand whether it is datasource bug or backend bug.

arturminchukov:
I think it is mostly related to the VL backend if it doesn't return the `_time` field in the log line (I'm not sure if it can happen in the real environment). In this case, the datasource backend could be down with the error.

It was fixed during the refactoring in commit a07be73.

@immanuwell, thanks for the contribution.
Could you please just leave the test case and changelog in this pr?

arturminchukov (approved):
LGTM! @immanuwell thanks for the contribution!

PR #673, opened 2026-06-17
Show stream labels in context modal

### LINKED ISSUE #665: Show stream labels in context modal

arturminchukov:
Related issue: #665 

### Describe Your Changes

1. Add parsing streams (was removed in the last commit) on the backend part to set streamId and stream labels to custom metadata.
2. Moved the code related to Context to a separate class. 
3. Added stream labels to context modal. Now it is possible to widen stream fields to view logs from other pods/services.
image

Context for all pods: 
image

### Checklist

---
## Summary by cubic
Show stream labels in the “Show context” modal so you can widen context by toggling labels and view logs across pods/services. Backend parses `_stream` and sends per‑row labels/ids via `meta.custom` to power the selector. Addresses #665.

- **New Features**
  - Added a stream label selector to the context modal; each label is a toggleable chip.
  - Context query uses `_stream_id` by default, or `_stream:{...}` from enabled labels when some are toggled off.
  - Safeguards: at least one label stays enabled; label keys with spaces are quoted; values are escaped.
  - Works for both instant and streaming responses; supports Explore and Dashboards.

- **Refactors**
  - Backend: parse `_stream` per row into `meta.custom.streams` and `meta.custom.streamIds`; remove `_stream` from log labels.
  - Frontend: moved context logic into a new `LogContextProvider`; added `LogContextUI` and a shared `ToggleChip` component.
  - Kept `meta.custom` intact in frame processing, passing through `streams/streamIds`.
  - Expanded tests for response parsing, stream selector parsing (incl. invalid cases), frame processing, and the new provider/UI.

<sup>Written for commit f17238a1f2ad8296826abc545f3601e8e1da50da. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VictoriaMetrics/victorialogs-datasource/pull/673?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pkg/plugin/response_logs.go
```diff
@@ -113,7 +127,17 @@ func parseInstantResponse(reader io.Reader) backend.DataResponse {
...
 		}
 
-		// not deleting `_stream` field from `value`, cause it's needed for displaying stream labels in the log context
+		// parse `_stream` into a per-row label map for the log context UI and
+		// drop it from the row so it does not show up among the log labels
+		rowStream := string(value.GetStringBytes(streamField))
+		value.Del(streamField)
+		stf, err := utils.ParseStreamFields(rowStream)
+		if err != nil {
+			return newResponseError(fmt.Errorf("error parse _stream field: %s", err), backend.StatusInternal)
```
Loori-R: Invalid `_stream` now fails the whole logs response. Since it is only needed for the context selector, can we skip `meta.custom.streams` for this row instead?

pkg/plugin/response_logs.go
```diff
@@ -113,7 +127,17 @@ func parseInstantResponse(reader io.Reader) backend.DataResponse {
 			rowStreamId = value.GetStringBytes(streamIdField)
 		}
 
-		// not deleting `_stream` field from `value`, cause it's needed for displaying stream labels in the log context
+		// parse `_stream` into a per-row label map for the log context UI and
+		// drop it from the row so it does not show up among the log labels
+		rowStream := string(value.GetStringBytes(streamField))
+		value.Del(streamField)
```
Loori-R: Does this code remove the `_stream` field from the response? If yes, could this break existing Explore/table views, field overrides, or Grafana transformations that reference `_stream`?

src/logContext/LogContextProvider.tsx
```diff
@@ -0,0 +1,177 @@
...
+  constructor(private readonly datasource: VictoriaLogsDatasource) {}
+
+  getRowStreamId = (row: LogRowModel): string => {
+    const streamIds = row.dataFrame.meta?.custom?.streamIds;
+    if (streamIds && streamIds.length > 0 && streamIds[row.rowIndex]) {
+      return streamIds[row.rowIndex];
+    }
+
+    return '';
+  };
```
Loori-R, line 56: The previous `_stream_id` fallback from `row.labels` was removed here. Is that intentional? If `meta.custom.streamIds` is missing, the context query now uses `_stream_id:""`.

Loori-R (approved):
Looks good overall. I left a few questions.

pkg/plugin/response_logs.go
arturminchukov: It shouldn't affect the backward compatibility, as in the current release, this field is also removed on the backend side.

pkg/plugin/response_logs.go
arturminchukov: I just partially reverted the changes as it was early - https://github.com/VictoriaMetrics/victorialogs-datasource/blob/v0.28.0/pkg/plugin/response.go#L122 . We also threw an error. 
IMO, the `_stream` field should always be in the log. So if something is wrong with the `_stream` field, it will be better to throw an error. So it could be fixed.

src/logContext/LogContextProvider.tsx
arturminchukov: Good point! 
But I think that from the DS backend we should get `streamId` every time and need to move this logic to the DS backend.
I thought that `_stream_id` should always be. But also need to handle edge cases correctly where the user manually removed system fields (`_stream`, `_stream_id`) via a query.

arturminchukov:
Added handler for cases:
1. no `_stream_id` and no `_stream`
2. no `_stream_id`
3. no `_stream`
if user manually deletes these fields, e.g. `* | delete _stream, _stream_id`.

image
image

Issue #519, opened 2025-12-23
support highlighting searched keywords?


devopslyf:
### Describe the bug

https://grafana.com/docs/grafana/latest/visualizations/explore/logs-integration/#highlight-searched-words

### To Reproduce

support highlighting searched keywords?

### Versions of VictoriaLogs datasource and VictoriaLogs backend

v1.31.0

PR #675, opened 2026-06-18
feat(logs): highlight LogsQL search terms in log lines (#519)

arturminchukov:
Related issue: #519 

### Describe Your Changes

Added highlight LogsQL search terms in log lines for cases:
1. `msgValue ...` , `"msgValue"` 
2. `_msg: msgValue`, `_msg: "msgValue"`
3. for pipes filter : ` * | filter msgValue`, ` * | filter "msgValue"`, ` * | "msgValue"`, ` * | field: value _msg: "msgValue"`

The search words are set to the custom metadata `searchWords` in [this line](https://github.com/VictoriaMetrics/victorialogs-datasource/blob/d452d8c457023225adbaca6b1755fd2698d9ff8f/src/transformers/frameProcessors/streamFrameProcessor.ts).

image

### Checklist

### DIFF
--- a/src/LogsQL/extractMsgSearchWords.ts
+++ b/src/LogsQL/extractMsgSearchWords.ts
@@ -3,6 +3,7 @@ import { escapeRegExp } from 'lodash';
 import { skipBalanced } from '../utils';
 
 import { splitByPipes } from './splitByPipes';
+import { stripComments } from './stripComments';
 
 // Tabs/newlines are normalized to spaces in extractMsgSearchWords, so a plain space is enough here
 const TERM_SEPARATORS = [' ', ':', '|', '(', ')', '{', '}'];
@@ -204,9 +205,33 @@ function scanFilterSegment(segment: string): string[] {
       while (segment[i] === ' ') {
         i++; // a space after the colon still binds the value to its field
       }
+      // grouped value: `field:(...)` — the whole group binds to the field
+      if (segment[i] === '(') {
+        if (word === '_msg') {
+          continue; // scan the group as default _msg context
+        }
+        i = skipBalanced(segment, i, '(', ')'); // group filters a non-_msg field — skip it
+        negateNext = false;
+        continue;
+      }
       const { term, next } = readValueTerm(segment, i);
       i = next;
+      // function-style value: `field:fn(...)` — the value word is a function name, not a term
+      if (segment[i] === '(') {
+        if (word === '_msg') {
+          continue; // scan the function args as default _msg context
+        }
+        i = skipBalanced(segment, i, '(', ')'); // function filters a non-_msg field — skip it
+        negateNext = false;
+        continue;
+      }
       emit(word === '_msg' ? term : null);
+      continue;
+    }
+
+    // function-style filter call: `word(...)` — the word is a function name, not a term
+    // (e.g. i(error), exact(error)); the group itself is scanned as default _msg context
+    if (segment[i] === '(') {
       continue;
     }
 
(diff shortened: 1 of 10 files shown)


src/LogsQL/extractMsgSearchWords.ts
```diff
@@ -0,0 +1,272 @@
...
+      i = next;
+      emit(word === '_msg' ? term : null);
+      continue;
+    }
+
+    // bare word filter on _msg
+    if (word.endsWith('*')) {
+      word = word.slice(0, -1);
+    }
+    emit(word ? escapeRegExp(word) : null);
```
Loori-R, line 242: For query `i(error)`, the scanner returns both `i` and `error` in `searchWords`.

This may highlight every `i` in log lines. We should avoid adding the filter function name to `searchWords`, or consider skipping too-short search terms.

See [case-insensitive filter](https://docs.victoriametrics.com/victorialogs/logsql/#case-insensitive-filter).

src/LogsQL/extractMsgSearchWords.ts
```diff
@@ -0,0 +1,272 @@
...
+    if (segment[i] === ':') {
+      i++; // consume ':'
+      while (segment[i] === ' ') {
+        i++; // a space after the colon still binds the value to its field
+      }
+      const { term, next } = readValueTerm(segment, i);
+      i = next;
+      emit(word === '_msg' ? term : null);
+      continue;
+    }
```
Loori-R, line 230: For query `app:(buggy_app OR foobar)`, these words are filters for the `app` field, not for `_msg`.

But `readValueTerm` stops at `(`, so the scanner continues inside the group and adds `buggy_app` and `foobar` to `searchWords`.

Should we skip the whole grouped value after a non-`_msg` field?

src/LogsQL/extractMsgSearchWords.ts
```diff
@@ -0,0 +1,272 @@
...
+/**
+ * Extracts regex-ready search terms targeting the default `_msg` field from a LogsQL query,
+ * for use in Grafana `frame.meta.searchWords` highlighting. The first segment (before the first
+ * top-level `|`) is always scanned as a filter; later segments are scanned only when they are
+ * filter pipes (start with `filter`, a quoted phrase, or `_msg:`). Literal terms are escaped;
+ * regexp filters are returned raw.
+ */
+export function extractMsgSearchWords(expr = ''): string[] {
+  // Normalize tabs/newlines to spaces up front so the scanner only deals with one whitespace kind
+  const segments = splitByPipes(expr.replace(/[\t\n\r]/g, ' '));
```
Loori-R: Comments are not stripped before extracting `searchWords`.

For query `error # debug`, the scanner may add `debug` to `searchWords`, even though it is only a comment.

Should we remove comments before normalizing newlines?

See [LogsQL comments](https://docs.victoriametrics.com/victorialogs/logsql/#comments).

Loori-R (commented):
Left a few comments for edge cases where `searchWords` can be extracted incorrectly.

arturminchukov:
Added processing of edge cases of extracting the _msg values. 
image

dmitryk-dk (approved):
LGTM!

src/LogsQL/extractMsgSearchWords.ts
```diff
@@ -0,0 +1,298 @@
...
+        i = skipBalanced(segment, i, '(', ')'); // group filters a non-_msg field — skip it
+        negateNext = false;
+        continue;
+      }
+      const { term, next } = readValueTerm(segment, i);
+      i = next;
+      // function-style value: `field:fn(...)` — the value word is a function name, not a term
+      if (segment[i] === '(') {
+        if (word === '_msg') {
+          continue; // scan the function args as default _msg context
```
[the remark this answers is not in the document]
arturminchukov: Negation isn't lost here. In the _msg:(...) / _msg:fn(...) branches the continue fires while i still points at ( and negateNext is intentionally kept. The next iteration hits the grouping-paren handler, which skips the whole group via skipBalanced when negateNext is set — so inner terms aren't emitted. It's delegated there on purpose to avoid duplicating the skip logic.

You could check in the test files.

Loori-R (approved):

PR #676, opened 2026-06-19
feat(derivedFields): support linking to a VictoriaLogs datasource (logs → logs)

### LINKED ISSUE #671: Derived field internal link to a VictoriaLogs datasource opens Explore with empty query (logs->logs unsupported)
## What I'm trying to do

Make a log field clickable so it pivots to "all logs that share this id" in the same VictoriaLogs datasource. Concretely: a `request_id` (or `trace_id`) derived field that, on click, opens Explore running a LogsQL query like `*:${__value.raw}` against the same logs datasource. Same idea as the trace_id->Jaeger derived field, but logs->logs instead of logs->traces.

## What actually happens

If I set a derived field with `datasourceUid` pointing at a `victoriametrics-logs-datasource`, clicking it opens Explore on that datasource with an **empty** query. The LogsQL expr never gets populated, so the link is effectively dead.

## Why (had a look at the source)

In `src/transformers/fields/derivedField.ts` (checked on v0.28.0), the internal link is built like this:

```ts
internal: {
  query: { query: derivedFieldConfig.url, queryType: queryType(dsSettings?.type) },
  datasourceUid: derivedFieldConfig.datasourceUid,
  datasourceName: dsSettings?.name ?? 'Data source not found',
},
```

The query object uses the key `query`, which is what Jaeger/Zipkin/Tempo expect. But the VictoriaLogs `Query` type (`src/types.ts`) reads `expr`, not `query`:

```ts
export interface Query extends DataQuery {
  expr: string;
  ...
}
```

So when the target is another VictoriaLogs datasource, `{ query: "*:<id>" }` is dropped on the floor and Explore opens with no expr. The `queryType` switch also only knows `tempo` / `grafana-x-ray-datasource`, nothing for logs. That's why every derived-field example in the docs links out to a trace datasource and never back to logs.

## Suggested fix

When the resolved target datasource type is `victoriametrics-logs-datasource`, build the internal query as `{ expr: derivedFieldConfig.url, refId: 'A' }` instead of `{ query, queryType }`. That'd let logs->logs derived fields actually run a LogsQL query on click.

Happy to send a PR if you'd take it.

## Versions

- VL datasource: v0.28.0
- Grafana: 12.x

arturminchukov:
Related issue: #671 

### Describe Your Changes

Add VL DS to the allowed DS list in the derived fields configuration.
- Build the internal Explore query as `expr`/`refId` when the derived field target is a VictoriaLogs datasource, so LogsQL is actually populated; keep `query`/`queryType` for tracing datasources (Tempo, Jaeger, X-Ray)
- Allow selecting a VictoriaLogs datasource (alongside tracing ones) in the derived field datasource picker

Demo, how to show all logs related to `trace_id`: 

https://github.com/user-attachments/assets/06129ac2-006d-4e9a-a425-c9898e57f56d

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## tip
 
 * FEATURE: keep log fields in the order returned by VictoriaLogs (for example the order from `| fields a, b, c`) instead of sorting them alphabetically. See [#563](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/563).
+* FEATURE: derived fields can now link to another VictoriaLogs datasource (logs → logs). Clicking such a field opens Explore with a ready LogsQL query, letting you pivot from a value like `trace_id` or `request_id` to all related logs. See [#671](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/671).
 
 * BUGFIX: lower the default datasource `Maximum lines` limit from 1000 to 50 to reduce browser rendering load and prevent the page from freezing on heavy log results. You can change this value in the datasource configuration. See [Line limits](https://docs.victoriametrics.com/victorialogs/integrations/grafana/#line-limits) docs and [#669](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/669).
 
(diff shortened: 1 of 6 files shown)


Loori-R (approved):
LGTM!

PR #682, opened 2026-06-29
add a `:` (word filter) operator to Log Level Rules

### LINKED ISSUE #611: Support LogsQL word filter (:) in Log Level Rules
### Is your feature request related to a problem? Please describe

`Log Level Rules` in the VictoriaLogs datasource currently support only `=`, `!=`, `=~`, `<`, and `>`.

This makes it impossible to express some native LogsQL field filters in datasource rules, for example:

```
level:error
message:error
exception:timeout
```

Add a new rule type for LogsQL word filter semantics, mapped to `field:value`. The local rule evaluation should follow LogsQL word-filter semantics as closely as possible, not a naive substring match. For example, if the rule is:

```
field = message
operator = :
value = error
level = error
```

then the datasource should behave consistently with LogsQL for cases like:

- an error happened -> match
- error: cannot open file -> match
- multiple errors occurred -> should not match error
- terror detected -> should not match error

arturminchukov:
Related issue: #611 

### Describe Your Changes

Add a `:` (word filter) operator to Log Level Rules, matching a field value by whole word like LogsQL (e.g. `error` matches the word `error`, not `terror`)

### Checklist

### DIFF
--- a/src/configuration/LogLevelRules/const.ts
+++ b/src/configuration/LogLevelRules/const.ts
@@ -15,6 +15,7 @@ export const LOG_OPERATOR_OPTIONS: LogOperatorOption[] = [
   { label: '=~', description: 'Matches regex', value: LogLevelRuleType.Regex },
   { label: '<', description: 'Less than', value: LogLevelRuleType.LessThan },
   { label: '>', description: 'Greater than', value: LogLevelRuleType.GreaterThan },
+  { label: ':', description: 'Word filter', value: LogLevelRuleType.WordFilter },
 ];
 
 export const OperatorLabels: Record<LogLevelRuleType, string> = {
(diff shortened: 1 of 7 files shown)


dmitryk-dk (approved):

src/configuration/LogLevelRules/const.ts
```diff
@@ -33,6 +35,7 @@ export const OperatorLabelsQueryBuilder: Record<LogLevelRuleType, (rule: LogLeve
   [LogLevelRuleType.Regex]: (rule) => `${rule.field}:~"${rule.value}"`,
   [LogLevelRuleType.LessThan]: (rule) => `${rule.field}:<"${rule.value}"`,
   [LogLevelRuleType.GreaterThan]: (rule) => `${rule.field}:>"${rule.value}"`,
+  [LogLevelRuleType.WordFilter]: (rule) => `${rule.field}:"${rule.value}"`,
```
Loori-R, line 38: Could you check this edge case?

For this `WordFilter` rule, `value` is empty:

```ts
{
  field: 'message',
  operator: LogLevelRuleType.WordFilter,
  value: '',
  level: LogLevel.error,
}
```

Local evaluation returns no match, but the query builder still generates:

```logsql
message:""
```

So the same empty-value rule may behave differently between `extractLevelFromLabels()` and the generated LogsQL level expression.

Should we skip empty `wordFilter` rules here or prevent saving them?

arturminchukov: Good point! Fixed it.

Loori-R (approved):

PR #685, opened 2026-07-09
fix(raw logs): restore the `_stream` label in log query results

arturminchukov:
### Describe Your Changes

1. Restored the `_stream` field in both raw and live logs. Now it has the same behavior as the vmui.
2. Refactored the `response_log`, moved the common functionality to a separate function. 
3. Fixed tests by adding the missing `_stream` filter.

### Checklist

### DIFF
--- a/pkg/plugin/response_logs.go
+++ b/pkg/plugin/response_logs.go
@@ -34,8 +34,6 @@ func buildLogID(ts, msg, streamId []byte) string {
 	return strconv.FormatUint(h.Sum64(), 16)
 }
 
-var nowFunc = time.Now
-
 // streamFieldsToMap turns the parsed `_stream` fields into a label->value map
 // for meta.custom.streams, consumed by the "Show context" stream label selector
 func streamFieldsToMap(stf []utils.StreamField) map[string]string {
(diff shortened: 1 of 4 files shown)


dmitryk-dk (approved):
LGTM!

pkg/plugin/response_logs.go
```diff
@@ -46,10 +44,31 @@ func streamFieldsToMap(stf []utils.StreamField) map[string]string {
 	return m
 }
 
-// parseStreamResponse reads data from the reader and collects
-// fields and frame with necessary information
-func parseInstantResponse(reader io.Reader) backend.DataResponse {
+type LogRow struct {
```
hagen1778: what's the reason for LogRow and LogFrame to be public?

pkg/plugin/response_logs.go
```diff
@@ -62,10 +81,85 @@ func parseInstantResponse(reader io.Reader) backend.DataResponse {
...
-	streams := make([]map[string]string, 0)
+	return &LogFrame{
+		frame:     data.NewFrame("", timeFd, lineField, idField, labelsField),
+		streams:   make([]map[string]string, 0),
+		streamIds: make([]string, 0),
+	}
+}
+
+// parseLogLine trims, decodes and validates that the line is a JSON object.
+func parseLogLine(parser *fastjson.Parser, b []byte) (*fastjson.Value, error) {
```
hagen1778: I think this function name should be `parseJsonLine`, as it is what it does

pkg/plugin/response_logs.go
```diff
@@ -62,10 +81,85 @@ func parseInstantResponse(reader io.Reader) backend.DataResponse {
...
+	if value.Exists(streamField) {
+		rawStream := string(value.GetStringBytes(streamField))
+		stf, err := utils.ParseStreamFields(rawStream)
+		if err != nil {
+			return LogRow{}, fmt.Errorf("error parse _stream field: %s", err)
+		}
+		streamMap = streamFieldsToMap(stf)
+	}
+
+	value.Del(timeField)
```
hagen1778, line 137: Why there is need in deleting fields from the object?

hagen1778 (commented):
It would be great to mention in commit why this was rolled back, so it won't happen again in future.

pkg/plugin/response_logs.go
arturminchukov: We are deleting the `_time` and `_msg` because these fields exist in Grafana as separate fields. We add rest fields to the Grafana's `Labels` field.

pkg/plugin/response_logs.go
arturminchukov: No reason, changed it

arturminchukov:
> It would be great to mention in commit why this was rolled back, so it won't happen again in future.

We don't actually roll it back, cause deleting the `_stream` field was implemented a long time ago. We parse it to separate stream labels, so the query ` * | field_values _stream` worked. In the last release, I removed the parsing of this field, cause we have stream labels as separate fields, but left removing `_stream` labels. So it cause the empty result in query `* | field_values _stream`.

PR #686, opened 2026-07-09
fix(live tailing): pick up query changes during live tailing

arturminchukov:
### Describe Your Changes

Fixed live tailing. Previously, changing the query didn't affect the live tailing, cause we have the same channel path:
- Include a hash of the query in the Grafana Live channel path, so editing `expr` creates a new channel instead of reusing the old subscription that kept streaming the previous query until reload
- Add `hashString` utility (deterministic 7-char base36 hash) with tests

### Checklist

### DIFF
--- a/src/utils/string/hashString/hashString.ts
+++ b/src/utils/string/hashString/hashString.ts
@@ -1,14 +1,16 @@
 /**
- * Returns a deterministic, fixed-length 7-character base36 hash of a string.
+ * Returns a deterministic base36 hash of a string.
  *
- * Fast non-cryptographic hash (djb2-style via Math.imul), suitable for cache keys,
- * channel paths and identity checks. It is not collision-proof and not for security use.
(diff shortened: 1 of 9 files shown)


src/utils/string/hashString/hashString.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Returns a deterministic, fixed-length 7-character base36 hash of a string.
+ *
+ * Fast non-cryptographic hash (djb2-style via Math.imul), suitable for cache keys,
+ * channel paths and identity checks. It is not collision-proof and not for security use.
+ */
+export function hashString(str: string): string {
```
dmitryk-dk, line 8: nit: i think this function can be simplified. There are several possibilities

dmitryk-dk (approved):
LGTM!

arturminchukov:
Fixed channel closing from the datasource backend

PR #693, opened 2026-07-13
fix(highlighting): extract searchWords from the interpolated query (#684)

### LINKED ISSUE #684: Bug: searchWords for highlighting is extracted from raw query instead of interpolated query in VictoriaLogs datasource
### Describe the bug

In victoriametrics-logs-datasource, log highlighting (via searchWords metadata) does not work when using template variables (e.g., a textbox variable $search).

### To Reproduce

// Current implementation (concept):
const searchWords = getSearchWords(target.expr);

// Proposed fix:
const interpolatedQuery = this.templateSrv.replace(target.expr, options.scopedVars);
const searchWords = getSearchWords(interpolatedQuery);

### Versions of VictoriaMetrics datasource and VictoriaMetrics backend

Plugin version: v0.29
VictoriaMetrics version: v1.50

### Link to dashboard in VictoriaMetrics grafana playground

_No response_

### Please provide dashboard JSON if it is possible

_No response_

### Additional information

_No response_

arturminchukov:
Related issue: #684 

### Describe Your Changes

- Pass an explicit interpolateExpr callback from the datasource through transformBackendResult into the stream frame processor, so meta.searchWords is built from the query with template variables resolved
- Previously a query like `_msg:$search` highlighted the literal variable name instead of its value, so nothing matched in the log lines

### Checklist

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -174,7 +174,8 @@ export class VictoriaLogsDatasource
             response,
             fixedRequest,
             this.derivedFields ?? [],
-            this.getActiveLevelRules()
+            this.getActiveLevelRules(),
+            (expr) => this.interpolateString(expr, fixedRequest.scopedVars)
           )
         )
       );
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * BUGFIX: restore the `_stream` label in log query results. Previously, it was hidden from the log line labels. See [pr #685](https://github.com/VictoriaMetrics/victorialogs-datasource/pull/685).
 * BUGFIX: live tailing now picks up query changes immediately. Previously, editing the query expression while live logs were streaming kept showing results for the previous query until the page was reloaded.
 * BUGFIX: pad missing `_time` values in instant log query responses so mixed rows with and without `_time` still produce valid Grafana log frames. Thanks to @immanuwell for their contribution.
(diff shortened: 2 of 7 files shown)


src/datasource.ts
```diff
@@ -174,7 +174,8 @@ export class VictoriaLogsDatasource
             response,
             fixedRequest,
             this.derivedFields ?? [],
-            this.getActiveLevelRules()
+            this.getActiveLevelRules(),
+            (expr) => this.interpolateString(expr, fixedRequest.scopedVars)
```
Loori-R, line 178: This interpolation can diverge from the query actually executed for Grafana built-in variables.

Example:
- query: `_msg:$__interval`
- `scopedVars.__interval`: `20s`
- custom `step`: `5s`

The backend executes `_msg:5s`, while `meta.searchWords` may contain `20s`.

Could we resolve `$__interval` / `$__interval_ms` from the final `request.intervalMs` and add a test for this case?

Loori-R (approved):
LGTM, just one edge-case comment.

src/datasource.ts
arturminchukov: Thanks, the divergence is real, but I don't think it's worth fixing: it only shows up when a custom step is combined with $__interval inside a message filter, and searching log text for the panel's interval value is not a realistic use case. All practical uses of $__interval (stats bucketing, step expressions) sit behind pipes, where no `searchWords` are extracted, so highlighting is unaffected.

PR #698, opened 2026-07-20
fix(log context): stop duplicating lines and cap rows in "Show context" (#692)

### LINKED ISSUE #692: Grafana: lines within 1s after the selected row are displayed twice in 'show context'
### Describe the bug

In Grafana's "Show context" modal, every log line timestamped within **1 second after** the selected (anchor) line appears **twice**. Since logs of one request usually cluster within milliseconds, in practice the whole visible context is duplicated. Querying any of those lines directly returns a single copy — storage is fine, this is a fetch/display artifact.

What we found reading the source (plugin v0.29.0 + Grafana 13.0.1):

1. The plugin's context time ranges **overlap by 1 second**:

   ```ts
   // src/logContext/LogContextProvider.tsx (createContextTimeRange)
   const overlap = 1000;
   // Backward:  [t - 2h, t + overlap]
   // Forward:   [t,      t + 2h]
   ```

   So both direction queries return every line in `[t, t+1s]`. (The overlap presumably exists because the context request's `range.from/to` is millisecond-precision, and ending the backward range exactly at `t` would drop lines in the same millisecond — Loki's provider does that and same-instant lines get lost, see grafana/grafana#64966.)

2. On the Grafana side, the context modal's duplicate filter doesn't catch this: `LogRowContextModal` fires its above/below initial loads **concurrently**, and each load dedupes (`containsRow`: entry + `timeEpochNs`) against a stale snapshot containing only the anchor row — so lines fetched by both direction queries pass both filters and are appended twice.

Since the overlap lives in the plugin and the duplicate filtering lives in Grafana core, this probably needs to be looked at together — happy to provide more details or file a linked issue on the Grafana side if that helps.

### To Reproduce

1. Any VictoriaLogs datasource, query logs where several lines share the same second (e.g. one HTTP request's log lines).
2. Click **Show context** on one of them.
3. Lines just after the anchor are listed twice; lines further than 1s away appear once.

### Versions of VictoriaLogs datasource and VictoriaLogs backend

- Plugin version: v0.29.0
- VictoriaLogs version: v1.50.0 (single-node)
- Grafana version: v13.0.1+security-01
- Vmauth: not used (Grafana talks directly to the victoria-logs-single service)

### Link to dashboard in Victoria Metrics

https://play-grafana.victoriametrics.com/explore?schemaVersion=1&orgId=1&panes=%7B%22vls%22%3A%7B%22datasource%22%3A%22PD775F2863313E6C7%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%2A%22%2C%22datasource%22%3A%7B%22type%22%3A%22victoriametrics-logs-datasource%22%2C%22uid%22%3A%22PD775F2863313E6C7%22%7D%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-5m%22%2C%22to%22%3A%22now%22%7D%7D%7D

Open the link (Explore, `*` over the last 5m on the playground's VictoriaLogs datasource), pick a log line that has neighbors in the same second (the demo data logs land a few ms apart), click **Show context** → the lines within 1s of the anchor are listed twice (screenshot attached).

Note when double-checking storage on the playground: some demo pods (e.g. the kafka pod) are ingested into **two different streams**, so a global search for a line can legitimately return 2 copies with different `_stream_id`s. That's unrelated — the context modal scopes its query to the anchor's `_stream_id`, and within that single stream each line has exactly one copy (`_stream_id:<id> "<line>" | stats count()` → 1), yet is displayed twice in the modal.

### Please provide dashboard JSON if it is possible

_No response_

### Additional information

_No response_

arturminchukov:
Related issue: #692

### Describe Your Changes

- split the two context queries at the selected row's timestamp with nanosecond precision so the previous 1s overlap no longer shows lines twice
- add formatNanosEpochToISO to render the anchor timestamp as RFC3339 with nanosecond precision
- limit each direction query to the line count requested by the context modal (fallback 50) instead of returning the whole 2h scan window

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * BUGFIX: live tailing now picks up query changes immediately. Previously, editing the query expression while live logs were streaming kept showing results for the previous query until the page was reloaded.
 * BUGFIX: pad missing `_time` values in instant log query responses so mixed rows with and without `_time` still produce valid Grafana log frames. Thanks to @immanuwell for their contribution.
 * BUGFIX: highlight search terms in log lines when the query uses template variables (e.g. `_msg:$search`). Previously, highlighting looked for the literal variable name instead of its value, so nothing was highlighted. See [#684](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/684).
+* BUGFIX: stop showing log lines twice in the "Show context" modal. Previously, the two context queries overlapped by one second around the selected row, so every line within that window was displayed twice; the queries now split at the selected row's timestamp with nanosecond precision. See [#692](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/692).
 
 ## v0.29.0
 
--- a/src/logContext/LogContextProvider.test.ts
+++ b/src/logContext/LogContextProvider.test.ts
@@ -35,25 +35,26 @@ describe('LogContextProvider', () => {
       },
       labels: {},
       timeEpochMs: 1700000000000,
+      timeEpochNs: '1700000000000123456',
       ...overrides,
     }) as unknown as LogRowModel;
 
   it('builds context query by _stream_id when all stream labels are enabled', async () => {
     const query = await provider.getLogRowContextQuery(buildLogRow());
-    expect(query?.expr).toBe('_stream_id:"stream-id-1" | sort by (_time) desc');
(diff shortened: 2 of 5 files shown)


Loori-R (approved):
LGTM!

PR #702, opened 2026-07-22
691 detect the stream filters due filter inout option

### LINKED ISSUE #691: Detect the stream filters due filter in/out option
### Is your feature request related to a problem?

When a user clicks **Filter for value / Filter out value** in the log details view (Explore / Logs panel), the datasource always adds a regular field filter:

```logsql
field:="value"
```

Regular field filters scan data blocks, so on large volumes such queries can be slow — even when the clicked field is actually a **stream field** (e.g. `app`, `namespace`, `instance`).

Detect whether the clicked field is a stream field and, if so, add it as a [stream filter](https://docs.victoriametrics.com/victorialogs/logsql/#stream-filter) instead:

```logsql
{field="value"}
```

Stream filters are resolved via the stream index and are the most efficient filter type in LogsQL, so this should noticeably reduce query execution time for filter in/out actions on stream fields.

### Implementation notes

- Detection can be **synchronous and free**: `toggleQueryFilter` receives `ToggleFilterAction` with an optional `frame: DataFrame`, and the backend already attaches per-row stream label maps to every logs frame (`meta.custom.streams`, built from the `_stream` field). The clicked key is a stream field iff it is present in those maps. This is the same mechanism Loki uses to detect label types.
- The filter chip can be marked with a flag (similar to the existing editor-only `fromLevelFilter` flag), and the serialization branch added in `getFilterInsertValue`, which already special-cases `_stream` / `_stream_id`. Both delivery paths (root-query prefix and `extra_filters`) accept LogsQL filters, so the stream syntax works in both.
- Operators map 1:1 — `{...}` supports `=`, `!=`, `=~`, `!~`.
- Fallback: if no frame is available in the action, keep the current behavior (regular field filter).

### Things to verify

- **Filter-out semantics:** `{k!="v"}` may also match streams *without* the label `k` (PromQL-like semantics). Compare with the current `k:!="v"` behavior and pick the equivalent form (possibly `-{k="v"}`).
- Stream filters passed via `extra_filters` are optimized by the index the same way as inline ones.

arturminchukov:
Related issue: #691 

### Describe Your Changes

1. Change storage for `streams` and `streamId` from `frame.custom.meta` to hidden fields `streams` and `streamId`. It allows us to get access to streams when we do `filter in` / `filter out`, cause in this function we have access only to log data, not custom.meta.
2. Fix the problem with multiple values in the dashboard while we navigate with multiple value variables to the Explore page.
3. Unify stream filters and adhoc filter chips; now, it looks similar (stream filters have a stream badge):
image

4. Fix the interpolation variables while navigating to the Explore page if builder mode is on. Previously, the variable wasn't interpolated into values in builder mode.
5. In `filter in/filter out` operations, detect if the field is a stream or not. Then add this field to the detected group.

### Checklist


src/datasource.ts
```diff
@@ -294,7 +301,14 @@ export class VictoriaLogsDatasource
...
         adHocFilters: chips,
-        extraStreamFilters: this.getExtraStreamFilters(query.streamFilters, scopedVars),
+        // The dashboard variables don't exist in Explore, so the editor state that
+        // regenerates the expression (builder model, stream filter values) must be
+        // interpolated too — otherwise raw `$var` placeholders leak back into the
+        // query on the first editor interaction
+        templateBuilder: query.templateBuilder
+          ? interpolateTemplateBuilder(query.templateBuilder, interpolate)
+          : query.templateBuilder,
+        streamFilters: query.streamFilters?.map((f) => ({ ...f, values: f.values.map(interpolate) })),
```
Loori-R: Could you please verify this with a multi-select variable test? For example, when `$app` resolves to `["api", "worker"]`, `streamFilters` should preserve two separate values and `applyTemplateVariables()` should produce `_stream:{app in ("api", "worker")}`, rather than `_stream:{app in ("api OR worker")}`.

src/datasource.ts
```diff
@@ -187,19 +190,19 @@ export class VictoriaLogsDatasource
...
 
-    if ((isFilterFor && !exists) || isFilterOut) {
-      next = [...next, { key, value, operator: isFilterFor ? '=' : '!=' }];
+    // Stream fields go to the stream filters (extra_stream_filters), resolved
+    // via the stream index — the fastest filter type in LogsQL. The value is
+    // stored raw: the stream filter serializer escapes it itself
+    if (frameHasStreamField(filter.frame, key)) {
+      const next = toggleStreamFilterValue(query.streamFilters ?? [], filter.type, key, filter.options.value);
+      const expr = query.expr || (next.length ? '*' : '');
+      return { ...query, expr, streamFilters: next.length ? next : undefined };
```
Loori-R: Could we also remove a matching value from `adHocFilters` when routing it to `streamFilters`? Otherwise a saved filter such as `app="api"` can remain active while “Filter out” adds `_stream:{app not_in ("api")}`, resulting in contradictory filters and no results.

src/components/QueryEditor/StreamFilters/streamFilterUtils.ts
```diff
@@ -32,7 +33,7 @@ export function streamFilterToString(filter: StreamFilterState): string {
     return '';
   }
 
-  const op = filter.operator || 'in';
+  const op = streamFilterOperator(filter);
 
   const valuesList = filter.values.map(formatStreamValue).join(', ');
   return `_stream:{${filter.label} ${op} (${valuesList})}`;
```
Loori-R: Should we wrap `filter.label` in quotes here? For example, `foo bar` would otherwise be serialized as `_stream:{foo bar in (...)}` instead of `_stream:{"foo bar" in (...)}`.

src/datasource.ts
arturminchukov: Thanks! fixed!

src/datasource.ts
arturminchukov: Thanks! fixed!

src/components/QueryEditor/StreamFilters/streamFilterUtils.ts
arturminchukov: Thanks! fixed!

arturminchukov:
The check failed due to package vulnerabilities. I will fix it in a separate pr, before the release.

Loori-R (approved):

PR #705, opened 2026-07-23
feat(logs volume): derive log levels server-side via format pipes

### LINKED ISSUE #700: use `format if ...` to optimize the hits query with `_msg` Log Level Rules
### Is your feature request related to a problem? Please describe

When the logs volume bar chart is loaded and **log level rules** reference the `_msg` field
(e.g. "if `_msg` contains `error` → level = error"), the plugin adds every rule field to the
`/select/logsql/hits` request as a `field=` arg:

```
/select/logsql/hits?query=<expr>&step=...&field=_msg&field=level
```

`field=_msg` makes VictoriaLogs group hit buckets by every distinct log message. Since `_msg`
is almost unique per log line, the response contains a separate `hits` series per unique
message. On busy log streams this results in a huge response payload and heavy client-side
aggregation: the frontend then applies the level rules to each returned series
(`aggregateRawLogsVolume` in `src/logsVolumeLegacy.ts`) just to collapse everything back
into ≤7 level series.

The same problem applies to any high-cardinality field used in level rules — `_msg` is just
the most common case.

Relevant code: `getSupplementaryQuery` in `src/datasource.ts` (builds the hits query with
`fields = [rule fields..., 'level']`).

Derive the log level **server-side** by appending conditional
[`format` pipes](https://docs.victoriametrics.com/victorialogs/logsql/#format-pipe) to the
query, and group hits by the derived field only:

```
<expr>
  | format "" as _level
  | format if (level:contains_common_case("err", "eror", "error") OR _msg:error) "error" as _level
  | format if ((level:contains_common_case("warn", "warning") OR _msg:warn) and _level:"") "warn" as _level
  ...
```

with a single grouping arg:

```
field=_level
```

`/select/logsql/hits` applies pipes by default (there is an explicit `ignore_pipes=1` arg to
opt out), see https://docs.victoriametrics.com/victorialogs/querying/#querying-hits-stats —
so no new endpoint is needed.

This collapses the response cardinality from "number of unique `_msg` values" down to
≤7 level series, and removes the client-side rule matching from the volume path entirely:
the `_level` label value maps directly to the level series.

The per-level LogsQL expressions already exist in the plugin — `buildLevelExprs`
(`src/utils/query/levelExpansion.ts`), currently used for ad-hoc `level` filter expansion
and query hints — so the `format` pipes can be generated from the same source without
duplicating the rule → LogsQL translation.

### Implementation notes / considerations

1. **Rule precedence.** Current frontend semantics: a valid `level` field value wins, then
   the *first* matching rule. Sequential `format` pipes are last-write-wins, so each
   subsequent pipe must be guarded with `and _level:""` (or the pipes emitted in reverse
   priority order) to preserve first-match-wins semantics.
2. **`unknown` level.** Rows matching no condition keep an empty `_level` — map the empty
   label value to `unknown` on the frontend instead of sending the expensive negation
   expression `!(<all levels OR'd>)`.
3. **Field name clash.** If ingested logs already contain a real `_level` field, unmatched
   rows would keep its original value. Resetting it first with an unconditional
   `| format "" as _level` (or choosing a more unique derived field name) avoids polluting
   the grouping.
4. **Escaping.** Rule values are interpolated into the generated filter expressions
   (`OperatorLabelsQueryBuilder`) — quotes/special characters in rule values must be escaped.
5. **`extra_filters` / `extra_stream_filters`** remain compatible: they are global
   constraints applied to the query independently of pipes.
6. **No rules configured.** Keep the current lightweight `field=level` request when there
   are no active level rules — pipes are only needed when rules exist.

### Describe alternatives you've considered

- `stats_query_range` with `... | stats by (_level) count()` — works, but unnecessary since
  `/select/logsql/hits` already applies pipes and returns data in the shape the volume
  panel consumes.
- Keeping client-side classification but capping the number of returned series — loses
  accuracy and still transfers a large payload.

arturminchukov:
Related issue: #700 

### Describe Your Changes

- Build conditional `format if` pipes from log level rules so /select/logsql/hits groups by a single derived field instead of every rule field — a `_msg` rule no longer produces a series per unique message value (#700)
- Strip any trailing sort pipe from the hits expression: VictoriaLogs hits returns empty values for fields created by pipes placed after `sort`, which would collapse the histogram into `unknown`
- Expand level chips with exact first-match-wins expressions (buildExactLevelExprMap) so chip filters select exactly the rows the histogram counted into each level
- Escape `"` and `\` in rule values interpolated into LogsQL (logsqlEscape), shared by filters and format pipes

### Checklist

### DIFF
--- a/src/configuration/LogLevelRules/const.ts
+++ b/src/configuration/LogLevelRules/const.ts
@@ -1,6 +1,8 @@
 import { LogLevel } from '@grafana/data';
 import { colors, ComboboxOption } from '@grafana/ui';
 
+import { quoteLogsQLValue } from '../../utils/query/logsqlEscape';
+
 import { LogLevelRule, LogLevelRuleType } from './types';
 
 export interface LogOperatorOption {
@@ -29,13 +31,13 @@ export const OperatorLabels: Record<LogLevelRuleType, string> = {
 };
 
 export const OperatorLabelsQueryBuilder: Record<LogLevelRuleType, (rule: LogLevelRule) => string> = {
-  [LogLevelRuleType.Equals]: (rule) => `${rule.field}:"${rule.value}"`,
-  [LogLevelRuleType.CaseInsensitiveEquals]: (rule) => `${rule.field}:contains_common_case("${rule.value}")`,
-  [LogLevelRuleType.NotEquals]: (rule) => `${rule.field}:!"${rule.value}"`,
-  [LogLevelRuleType.Regex]: (rule) => `${rule.field}:~"${rule.value}"`,
-  [LogLevelRuleType.LessThan]: (rule) => `${rule.field}:<"${rule.value}"`,
-  [LogLevelRuleType.GreaterThan]: (rule) => `${rule.field}:>"${rule.value}"`,
-  [LogLevelRuleType.WordFilter]: (rule) => `${rule.field}:"${rule.value}"`,
+  [LogLevelRuleType.Equals]: (rule) => `${rule.field}:${quoteLogsQLValue(rule.value)}`,
+  [LogLevelRuleType.CaseInsensitiveEquals]: (rule) => `${rule.field}:contains_common_case(${quoteLogsQLValue(rule.value)})`,
+  [LogLevelRuleType.NotEquals]: (rule) => `${rule.field}:!${quoteLogsQLValue(rule.value)}`,
+  [LogLevelRuleType.Regex]: (rule) => `${rule.field}:~${quoteLogsQLValue(rule.value)}`,
+  [LogLevelRuleType.LessThan]: (rule) => `${rule.field}:<${quoteLogsQLValue(rule.value)}`,
+  [LogLevelRuleType.GreaterThan]: (rule) => `${rule.field}:>${quoteLogsQLValue(rule.value)}`,
+  [LogLevelRuleType.WordFilter]: (rule) => `${rule.field}:${quoteLogsQLValue(rule.value)}`,
 };
 
 export const LOG_LEVEL_OPTIONS: Array<ComboboxOption<LogLevel>> = Array.from(
--- a/src/configuration/LogLevelRules/const.ts
+++ b/src/configuration/LogLevelRules/const.ts
@@ -2,7 +2,6 @@ import { LogLevel } from '@grafana/data';
 import { colors, ComboboxOption } from '@grafana/ui';
 
 import { quoteLogsQLFieldName, quoteLogsQLValue } from '../../utils/query/logsqlEscape';
-import { capitalize } from '../../utils/string';
 
 import { LogLevelRule, LogLevelRuleType } from './types';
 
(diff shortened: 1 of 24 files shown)


src/configuration/LogLevelRules/const.ts
```diff
@@ -29,13 +31,13 @@ export const OperatorLabels: Record<LogLevelRuleType, string> = {
...
 
 export const OperatorLabelsQueryBuilder: Record<LogLevelRuleType, (rule: LogLevelRule) => string> = {
-  [LogLevelRuleType.Equals]: (rule) => `${rule.field}:"${rule.value}"`,
-  [LogLevelRuleType.CaseInsensitiveEquals]: (rule) => `${rule.field}:contains_common_case("${rule.value}")`,
-  [LogLevelRuleType.NotEquals]: (rule) => `${rule.field}:!"${rule.value}"`,
-  [LogLevelRuleType.Regex]: (rule) => `${rule.field}:~"${rule.value}"`,
-  [LogLevelRuleType.LessThan]: (rule) => `${rule.field}:<"${rule.value}"`,
-  [LogLevelRuleType.GreaterThan]: (rule) => `${rule.field}:>"${rule.value}"`,
-  [LogLevelRuleType.WordFilter]: (rule) => `${rule.field}:"${rule.value}"`,
+  [LogLevelRuleType.Equals]: (rule) => `${rule.field}:${quoteLogsQLValue(rule.value)}`,
```
Loori-R: `OperatorLabelsQueryBuilder` should preserve the same semantics as [ruleMatchers](https://github.com/VictoriaMetrics/victorialogs-datasource/blob/9f5a9952433dc2f34430d7faec62cc247e776d06/src/configuration/LogLevelRules/utils.ts#L36).

For example, the client-side `Equals` matcher uses strict equality:

```ts
fieldValue === rule.value
```

but the generated LogsQL currently uses:

```logsql
field:"value"
```

which is a word/phrase filter rather than an exact value comparison. This can classify values such as `"api server"` as matching an `Equals "api"` rule.

Should this use the exact LogsQL operator instead?

```suggestion
  [LogLevelRuleType.Equals]: (rule) => `${rule.field}:=${quoteLogsQLValue(rule.value)}`,
```

It may also be worth checking the other operators to ensure `OperatorLabelsQueryBuilder` and `ruleMatchers` remain semantically aligned.

src/configuration/LogLevelRules/const.ts
```diff
@@ -29,13 +31,13 @@ export const OperatorLabels: Record<LogLevelRuleType, string> = {
...
+  [LogLevelRuleType.CaseInsensitiveEquals]: (rule) => `${rule.field}:contains_common_case(${quoteLogsQLValue(rule.value)})`,
+  [LogLevelRuleType.NotEquals]: (rule) => `${rule.field}:!${quoteLogsQLValue(rule.value)}`,
+  [LogLevelRuleType.Regex]: (rule) => `${rule.field}:~${quoteLogsQLValue(rule.value)}`,
+  [LogLevelRuleType.LessThan]: (rule) => `${rule.field}:<${quoteLogsQLValue(rule.value)}`,
+  [LogLevelRuleType.GreaterThan]: (rule) => `${rule.field}:>${quoteLogsQLValue(rule.value)}`,
+  [LogLevelRuleType.WordFilter]: (rule) => `${rule.field}:${quoteLogsQLValue(rule.value)}`,
```
Loori-R: Should field names be escaped or quoted here as well?

For example, a field such as `log:level` would currently produce:

```logsql
log:level:="error"
```

instead of:

```logsql
"log:level":="error"
```

src/configuration/LogLevelRules/const.ts
```diff
@@ -29,13 +32,13 @@ export const OperatorLabels: Record<LogLevelRuleType, string> = {
...
 export const OperatorLabelsQueryBuilder: Record<LogLevelRuleType, (rule: LogLevelRule) => string> = {
-  [LogLevelRuleType.Equals]: (rule) => `${rule.field}:"${rule.value}"`,
-  [LogLevelRuleType.CaseInsensitiveEquals]: (rule) => `${rule.field}:contains_common_case("${rule.value}")`,
-  [LogLevelRuleType.NotEquals]: (rule) => `${rule.field}:!"${rule.value}"`,
-  [LogLevelRuleType.Regex]: (rule) => `${rule.field}:~"${rule.value}"`,
-  [LogLevelRuleType.LessThan]: (rule) => `${rule.field}:<"${rule.value}"`,
-  [LogLevelRuleType.GreaterThan]: (rule) => `${rule.field}:>"${rule.value}"`,
-  [LogLevelRuleType.WordFilter]: (rule) => `${rule.field}:"${rule.value}"`,
+  [LogLevelRuleType.Equals]: (rule) => `${quoteLogsQLFieldName(rule.field)}:=${quoteLogsQLValue(rule.value)}`,
+  [LogLevelRuleType.CaseInsensitiveEquals]: (rule) => `${quoteLogsQLFieldName(rule.field)}:contains_common_case(${quoteLogsQLValue(capitalize(String(rule.value)))})`,
```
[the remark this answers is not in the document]
arturminchukov: It is used only in the preset where this expression gives the necessary combinations of level words. e.g., `severity_text: contains_common_case("Error")` will be processed to `severity_text:in('Error', "ERROR", "error")`.

arturminchukov:
Check failed due to package vulnerabilities.

Loori-R (approved):
LGTM!

PR #707, opened 2026-07-30
fix(logs): deduplicate log row ids to stop lines overlapping (#706)

### LINKED ISSUE #706: Explore Logs: log lines render overlapping / stacked on top of each other after re-running a query
## What happened?

In Explore,  log lines sometimes render **visually overlapping / stacked on top of each other** after changing the query and running it again. The text of several log lines is drawn in the same vertical space, making the list unreadable.

The dataset itself is fine — the same response renders correctly after a fresh page load, so this looks like stale virtualization state (cached row heights/offsets) surviving a data change in the new `LogList` panel.

A similar report exists for 13.0.1: https://community.grafana.com/t/explore-panel-logs-are-stacked-on-each-other/163336.

## What did you expect to happen?

Every log line is rendered in its own row, with no visual overlap, regardless of how many times the query is edited and re-run.

## How to reproduce

Probably, it appears when you modify the query or load more logs via infinite scroll.

**Actual:** after a re-run, some log lines are drawn on top of each other.

**Expected:** the list re-lays out cleanly for the new result set.

Image

## Environment

- Grafana: 12.4.3 - security2a
- Datasources: v0.29.0
- Browser: Chrome, firefox

arturminchukov:
Related issue: #706 

### Description

The problem was in the same LogID for the logs with the same time, stream_id, and _msg. To fix it:
- Include row labels in the log ID hash, so rows that differ only in their fields (e.g., trace_id) no longer share one ID
- Add idBuilder that suffixes repeated ids with _1, _2, ... within one response or live-tail session; the seen-set is size-capped, so long tail sessions don't grow memory unboundedly.

### Checklist

### DIFF
--- a/pkg/plugin/log_reader.go
+++ b/pkg/plugin/log_reader.go
@@ -31,7 +31,7 @@ func newLogReader(reader io.Reader) *logReader {
 	}
 }
 
-func (lr *logReader) ReadRow() (logRow, error) {
+func (lr *logReader) readRow() (logRow, error) {
 	for {
 		b, err := lr.readLine()
 		if errors.Is(err, errLineTooLong) {
--- a/pkg/plugin/log_reader.go
+++ b/pkg/plugin/log_reader.go
@@ -1,0 +1,155 @@
+package plugin
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/VictoriaMetrics/victorialogs-datasource/pkg/utils"
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/valyala/fastjson"
+)
+
+var errLineTooLong = errors.New("line too long")
+
+type logReader struct {
+	parser       fastjson.Parser
+	idb          *idBuilder
+	bufferReader *bufio.Reader
+	lineNum      int64
+}
+
+func newLogReader(reader io.Reader) *logReader {
+	bufferReader := bufio.NewReaderSize(reader, 64*1024)
+	return &logReader{
+		idb:          newIDBuilder(),
+		bufferReader: bufferReader,
+	}
+}
+
+func (lr *logReader) ReadRow() (logRow, error) {
+	for {
+		b, err := lr.readLine()
+		if errors.Is(err, errLineTooLong) {
+			backend.Logger.Debug("skipping line: line too long", "lineNumber", lr.lineNum)
+			continue
+		}
+		if err != nil {
+			return logRow{}, err
+		}
+		if len(bytes.TrimSpace(b)) == 0 {
+			continue
+		}
+		return lr.parseRow(b)
+	}
+}
+
+func (lr *logReader) readLine() ([]byte, error) {
+	lr.lineNum++
+	b, err := lr.bufferReader.ReadBytes('\n')
+	switch {
+	case err == nil:
+		return b, nil
+	case errors.Is(err, bufio.ErrBufferFull):
+		for errors.Is(err, bufio.ErrBufferFull) {
+			_, err = lr.bufferReader.ReadBytes('\n')
+		}
+		if err != nil && !errors.Is(err, io.EOF) {
+			return nil, fmt.Errorf("cannot read line in response: %w", err)
+		}
+		return nil, errLineTooLong
+	case errors.Is(err, io.EOF):
+		if len(b) > 0 {
+			return b, nil
+		}
+		return nil, io.EOF
+	default:
+		return nil, fmt.Errorf("cannot read line in response: %w", err)
+	}
+}
+
+// parseJsonLine trims, decodes and validates that the line is a JSON object.
+func (lr *logReader) parseJsonLine(b []byte) (*fastjson.Value, error) {
+	b = bytes.Trim(b, "\n")
+	value, err := lr.parser.ParseBytes(b)
+	if err != nil {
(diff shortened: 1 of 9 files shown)


pkg/plugin/logid.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// unique returns id for its first occurrence and id_N for the N-th repeat.
+// Suffixed ids cannot collide with base ids: base ids are hex digests and
+// never contain `_`. When a new id would grow the set beyond maxSize, the
+// set is reset first to keep memory bounded in long live-tail sessions.
+func (b *idBuilder) unique(id string) string {
+	n, ok := b.seen[id]
+	if !ok && len(b.seen) >= b.maxSize {
+		b.seen = make(map[string]int)
```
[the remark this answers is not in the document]
arturminchukov: Raised it to 100000. This amount of duplicated logs is enough for one query.

pkg/plugin/log_parser.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/VictoriaMetrics/victorialogs-datasource/pkg/utils"
+	"github.com/valyala/fastjson"
+)
+
+type logParser struct {
+	parser fastjson.Parser
+	idb    *idBuilder
+}
+
+func newLogParser() *logParser {
```
hagen1778: nit: 
1. make newLogParser() to accept `reader io.Reader`
2. initiate `bufio.NewReaderSize(reader, 64*1024)` when creating parser instance
3. remove forEachLogRow as it makes code more complicated for no good reason
4. Add method `logParser.ReadRow() (logRow, error)` and use it within `parseInstantResponse` and `parseStreamResponse`, calling this method inside their for-loops.

pkg/plugin/logid.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// idSeparator delimits the fields hashed into a log id so that values can't
+// run together across field boundaries (e.g. "ab"+"cd" vs "a"+"bcd").
+var idSeparator = []byte{0}
+
+func buildLogID(ts, msg, streamId, labels []byte) string {
+	h := fnv.New64a()
+	_, _ = h.Write(ts)
+	_, _ = h.Write(idSeparator)
```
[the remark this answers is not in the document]
arturminchukov: Not fixing this one: a boundary collision needs a NUL byte in a field adjacent to the shifted boundary, and no such input exists in practice. ts must parse as a timestamp (a NUL fails parsing and rejects the whole row), labels is re-marshaled JSON where fastjson escapes control chars back to \u0000 (no raw NUL bytes) and always starts with {, and _stream_id is a VictoriaLogs-generated hex string. The only field that can carry a raw NUL is _msg, and with both of its neighbors NUL-free the (x, y\0z) ↔ (x\0y, z) construction is impossible. Even if a base-id collision happened, the deduper still suffixes the repeat, so uids stay unique within a response/stream session — the residual risk is the same as the inherent 64-bit FNV collision probability (~1e-10 per response), which length-prefixing wouldn't remove either.

pkg/plugin/logid.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// unique returns id for its first occurrence and id_N for the N-th repeat.
+// Suffixed ids cannot collide with base ids: base ids are hex digests and
+// never contain `_`. When a new id would grow the set beyond maxSize, the
+// set is reset first to keep memory bounded in long live-tail sessions.
+func (b *idBuilder) unique(id string) string {
+	n, ok := b.seen[id]
+	if !ok && len(b.seen) >= b.maxSize {
+		b.seen = make(map[string]int)
```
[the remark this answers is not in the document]
arturminchukov: Leaving as is. The instant path can't hit the cap (maxLines ≤ 10 000), so a reset is only possible in live tail — and there Grafana keeps just the last 1000 rows (StreamingDataFrame default maxLength). A duplicate uid only matters while both copies are in that buffer, which rules out re-delivery and leaves one case: identical rows straddling the reset boundary within the same 1000-row window, once per 100 000 distinct ids. That's a transient render glitch that self-heals as the buffer rotates — not worth changing the id shape for. Will revisit with a generation marker on post-reset ids if it ever shows up in practice.

pkg/plugin/log_reader.go
```diff
@@ -0,0 +1,155 @@
...
+
+func newLogReader(reader io.Reader) *logReader {
+	bufferReader := bufio.NewReaderSize(reader, 64*1024)
+	return &logReader{
+		idb:          newIDBuilder(),
+		bufferReader: bufferReader,
+	}
+}
+
+func (lr *logReader) ReadRow() (logRow, error) {
```
hagen1778: No need to make it public

pkg/plugin/log_reader_test.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"encoding/json"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/VictoriaMetrics/victorialogs-datasource/pkg/utils"
+)
+
+func Test_logReader_parseRow(t *testing.T) {
```
hagen1778, line 12: How is this test different to `Test_logParser_parseRow`?

hagen1778 (approved):

pkg/plugin/log_reader_test.go
arturminchukov: I renamed logParser to logReader because I think this name better reflects the functionality `readRow`. So the test was also renamed. But the test is the same.

PR #708, opened 2026-08-03
fix: raise default line limit from 50 to 1000 to reduce load-more loss

arturminchukov:
### Describe Your Changes

Why: with infinite scroll ("load more"), rows sharing the boundary timestamp can be silently lost at every chunk junction. Grafana narrows the next chunk's `to` to the oldest loaded row's timestamp truncated to milliseconds (the wire format is ms), and VictoriaLogs treats `end` as exclusive ([start, end)), so boundary-millisecond rows that did not fit into the previous chunk under the line limit can never be fetched by any later chunk. Reproduced deterministically: 120 rows on one millisecond with limit 50 lose 70 rows with no indication; the reference Loki datasource on Grafana 13.1 loses exactly the same rows, so the defect is systemic (panel ms-boundary + exclusive end + line limit), not specific to this plugin.

Raising the default limit from 50 to 1000 makes chunks 20x larger: there are far fewer load-more junctions, and each junction is far less likely to land inside a dense same-millisecond group, shrinking the expected loss (up to 1ms of logs per junction). This is mitigation, not a complete fix — the proper fix (aligning the range edge, e.g. end+1ms, or an upstream ns-precision range) is tracked separately.

### Checklist


Loori-R (approved):
LGTM!

hagen1778 (approved):

PR #713, opened 2026-08-06
fix: send an explicit end bound in the datasource health check

### LINKED ISSUE #712: Datasource health check omits `end`, so it 400s on any instance with -search.maxQueryTimeRange set
## What happens

Saving a datasource against a VictoriaLogs instance that sets
`-search.maxQueryTimeRange` fails the health check with:

```
got response code 400: too big time range selected:
[2026-08-05T23:51:28.749548662Z, 2262-04-11T23:47:16.854775807Z];
it cannot exceed -search.maxQueryTimeRange=720h0m0s
```

The datasource itself works perfectly once saved — only "Save & test" fails.

## Why

`checkHealthWithInstance` sets `start` but never sets `end`:

https://github.com/VictoriaMetrics/victorialogs-datasource/blob/main/pkg/plugin/datasource.go#L839-L843

```go
values := u.Query()
values.Set("query", "*")
values.Set("limit", "1")
values.Set("start", "-5m")
```

With no `end`, VictoriaLogs defaults the upper bound to max int64 nanoseconds
(`2262-04-11T23:47:16.854775807Z`), so the effective range is
`now-5m … year 2262`. Any instance with `-search.maxQueryTimeRange` set below
that — i.e. any instance that sets it at all — rejects the request with 400.

This came in with #599, which moved the health check onto the `query` endpoint.

## Reproduction

Against an instance started with `-search.maxQueryTimeRange=720h`:

```console
$ curl -s -o /dev/null -w '%{http_code}\n' \
    'http://vlselect:9428/select/logsql/query?query=*&limit=1&start=-5m'
400

$ curl -s -o /dev/null -w '%{http_code}\n' \
    'http://vlselect:9428/select/logsql/query?query=*&limit=1&start=-5m&end=now'
200
```

## Suggested fix

Set an explicit upper bound in the health check:

```go
values.Set("start", "-5m")
values.Set("end", "now")
```

Every other query path already sends both bounds
(`pkg/plugin/query.go` lines 168-169, 232-233, 270-271), which is why only the
health check is affected.

## Workaround considered and rejected

Adding `?end=now` to the datasource URL does fix the health check, because
`checkHealthWithInstance` seeds its params from the base URL. But
`queryTailURL` also inherits base-URL params and only overrides `query`, so the
bound leaks into the live tail endpoint and truncates the stream. Not viable.

## Environment

- Plugin: v0.30.1 (code path unchanged on `main` as of filing)
- Grafana: 13.2.0
- VictoriaLogs: cluster mode, `-search.maxQueryTimeRange=30d`

dberkerdem:
Fixes #712.

## Problem

`checkHealthWithInstance` builds the health-check request with a `start` but no
`end`:

```go
values.Set("query", "*")
values.Set("limit", "1")
values.Set("start", "-5m")
```

With no `end`, VictoriaLogs defaults the upper bound to the maximum int64
nanosecond timestamp — `2262-04-11T23:47:16.854775807Z` — so the effective range
is `now-5m … year 2262`. Any instance running with `-search.maxQueryTimeRange`
therefore rejects the health check:

```
got response code 400: too big time range selected:
[2026-08-05T23:51:28.749548662Z, 2262-04-11T23:47:16.854775807Z];
it cannot exceed -search.maxQueryTimeRange=720h0m0s
```

The datasource works fine once saved — only "Save & test" fails — but it is
alarming, and it makes Grafana's `/api/datasources/uid/<uid>/health` endpoint
report the datasource as broken.

This has been present since #599 moved the health check onto the `query`
endpoint.

## Fix

Set an explicit upper bound:

```go
values.Set("start", "-5m")
values.Set("end", "now")
```

Every other query path already sends both bounds — `queryLogsURL`,
`statsQueryRangeURL` and `hitsQueryURL` all `Set("start", …)` and
`Set("end", …)` — so the health check was the only caller relying on the
server-side default.

## Verification

Against an instance started with `-search.maxQueryTimeRange=720h`:

```console
# before - what the plugin sends today
$ curl -s -o /dev/null -w '%{http_code}\n' \
    'http://vlselect:9428/select/logsql/query?query=*&limit=1&start=-5m'
400

# after - with the explicit end bound
$ curl -s -o /dev/null -w '%{http_code}\n' \
    'http://vlselect:9428/select/logsql/query?query=*&limit=1&start=-5m&end=now'
200
```

`gofmt` reports no changes.

## Note for reviewers

`end=now` keeps the health check's relative style, matching the existing
`start=-5m`. An absolute unix timestamp would work equally well if you prefer
consistency with the query paths, which format both bounds with
`strconv.FormatInt(...Unix(), 10)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## tip
 
-* BUGFIX: send an explicit `end` bound in the datasource health check. Without it, VictoriaLogs defaulted the upper bound to the maximum int64 nanosecond timestamp (year 2262), so instances running with `-search.maxQueryTimeRange` failed "Save & test" with `too big time range selected`. See [#712](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/712).
+* BUGFIX: send an explicit `end` bound in the datasource health check. Without it, VictoriaLogs defaulted the upper bound to the maximum int64 nanosecond timestamp (year 2262), so instances running with `-search.maxQueryTimeRange` failed "Save & test" with `too big time range selected`. See [#712](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/712). Thanks to @dberkerdem for contributing.
 
 ## v0.31.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## tip
 
+* BUGFIX: send an explicit `end` bound in the datasource health check. Without it, VictoriaLogs defaulted the upper bound to the maximum int64 nanosecond timestamp (year 2262), so instances running with `-search.maxQueryTimeRange` failed "Save & test" with `too big time range selected`. See [#712](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/712).
+
 ## v0.31.0
 
 * FEATURE: `Filter for value` / `Filter out value` in log details now adds the value to the `Stream filters` when the clicked field is a stream field. Stream filters are resolved via the stream index, so such filters run noticeably faster on large volumes. See [#691](https://github.com/VictoriaMetrics/victorialogs-datasource/issues/691).
--- a/pkg/plugin/datasource.go
+++ b/pkg/plugin/datasource.go
@@ -840,6 +840,10 @@ func checkHealthWithInstance(ctx context.Context, di *DatasourceInstance) (*back
 	values.Set("query", "*")
 	values.Set("limit", "1")
 	values.Set("start", "-5m")
+	// An explicit upper bound is required: without `end`, VictoriaLogs defaults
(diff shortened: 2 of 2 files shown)


arturminchukov (approved):
LGTM! @dberkerdem Thanks for the contributing.

arturminchukov:
@dberkerdem, could you please sing your commit? [Here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) is documentation on how to do it.

dberkerdem:
> @dberkerdem, could you please sing your commit? [Here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) is documentation on how to do it.

Done! @arturminchukov

